PR #26, opened 2024-02-21
Punctuation updates in README

chrisnager:
These changes remove periods from phrases and bulleted items in the root README file.


necolas:
Thanks, but the existing punctuation is preferred.

PR #29, opened 2024-02-23
Add babel plugin to optimize web output

### LINKED ISSUE #23: Compiler: inline wrapper component on web

Our plan is to try to inline the wrapper component on web, as it is relatively simple as long as some of its dependencies can refactored as importable code. For example, on web we'd take the source code:

```jsx
import { css, html } from 'react-strict-dom';

const styles = css.create({...});

<html.div
  dir="rtl"
  role="none"
  style={[styles.one, styles.two]}
/>

<html.label
  for={for}
  style={styles.label}
/>
```

And turn it into something like:

```jsx
import { css } from 'react-strict-dom';
import { privateStyles } from 'react-strict-dom/internals';

const styles = css.create({...});

<div
  dir="rtl"
  role="presentation"
  {...css.props(privateStyles.div, styles.one, styles.two)}
/>

<label
  dir="auto"
  htmlFor={for}
  {...css.props(privateStyles.label, styles.label}
/>
```

necolas:
The babel plugin can be used by web builds to inline the component wrapper and remove the runtime overhead.

Fix #23


packages/react-strict-dom/src/dom/runtime.js
```diff
@@ -0,0 +1,195 @@
...
+const strong: StrictReactDOMProps = {
+  style: [styles.inline, styles.strong]
+};
+const sub: StrictReactDOMProps = { style: styles.inline };
+const sup: StrictReactDOMProps = { style: styles.inline };
+const textarea: StrictReactDOMTextAreaProps = {
+  dir: 'auto',
+  style: [styles.inlineblock, styles.textarea]
+};
+const ul: StrictReactDOMProps = { style: styles.block };
```
vincentriemer, line 149: In the transformed output you're spreading the `defaultProps` but those contain the `style` prop which is invalid which is I'm guessing why you inject a `style={null}` prop before actually spreading the `stylex.props` call. If these definitions were separated into `defaultProps` and `defaultStyles` the transformed output could be simplified a bit to just something like `<div {...defaultProps.div} {...stylex.props(defaultStyles.div)} />`. I could be missing another reason why you took that approach though.

necolas: No you've understood correctly. It was just more code in the prod runtime to do it the way you mention. Another alternative would be to do away with `defaultProps` and have only `defaultStyles`, with the few elements that have a default prop (e.g., `dir=auto`) getting that prop supplied with some logic that is replicated in both the runtime wrapper and the babel plugin. That would probably be the most performant (fewest spreads in prod and least repeated code in bundles) approach, and the maintenance cost of duplication is probably worth it given we're unlikely to be adding/removing default props in any significant way going forward.

necolas: Pushed a diff that removes defaultProps but preserves the same result, so the transformed code looks like this instead `<div {...stylex.props(defaultStyles.div)} />` by default, and the 3 elements that have a default prop are special cased in the wrapper component and babel plugin

packages/react-strict-dom/babel.js
```diff
@@ -0,0 +1,159 @@
...
+        node.type === 'ImportDeclaration' && node.source.value === sourceValue
+    );
+  const findHtmlSpecifier = (specifiers) =>
+    specifiers.find((specifier) => specifier.imported.name === 'html');
+
+  return {
+    visitor: {
+      ImportDeclaration(path) {
+        if (path.node.source.value === packageName) {
+          const htmlSpecifier = findHtmlSpecifier(path.node.specifiers);
```
nmn, line 23: It's an edge-case, but a use may import `html` from `react-strict-dom` twice (with different aliases) within the same file.

In such an edge-case, this plugin would end up inserting duplicate imports for `defaultStyles`.

There is a second edge-case that the variable name `defaultStyles` may already be used within the file.

Both of these issues can be solved by using the `@babel/helper-module-imports` package to insert import statements instead.

nmn (approved):
Found some nits but they're edge-cases that can be fixed later.

---

But I wonder if it would better to make this a separate package?

packages/react-strict-dom/babel.js
necolas: Oh yeah, I noticed this edge case but forgot to account for it. Thanks for the pointer re: `@babel/helper-module-imports`

Issue #13, opened 2024-02-20
(feature) <input type="date"> element


necolas:
### Describe the feature request

Support the [`<input type="date">` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date). This element is not currently supported on native.

PR #31, opened 2024-02-23
Delete unused tag in examples app

TeTedo:
I see an issue that currently not support date type (#13 ).

But there is code line in App.js in exapmples app.

So reduce confusion for another user, delete unused data type input.

Thank you


necolas:
I don't think this is confusing anyone and I'd like to leave it there for when we get around to polyfilling it. Thanks

PR #35, opened 2024-02-24
Partial polyfill for CSS 'visibility'

necolas:
Implements a partial polyfill for non-inheritable 'visibility'. This does not polyfill the web behavior of being able to selectively override the 'visibility' of elements nested within an element with 'visibility:hidden'.

As result, this patch also removes 'visibility' from the inheritable text styles. Originally visibility was added here simply because it is considered an inheritable property on web, even though we didn't do anything with the property on native. In the context of these polyfills, we're only interested in being able to inherit text styles and cannot do anything with inherited visibility, so there is no functional change to the existing polyfill for inheritable text styles.

Also note that no attempt is made to implement 'collapse' as per W3C spec, as this behavior does not appear to be supported for flex items in Chrome anyway (which behave the same whether 'collapse' or 'hidden' is used.)

### DIFF
--- a/packages/react-strict-dom/src/native/stylex/index.js
+++ b/packages/react-strict-dom/src/native/stylex/index.js
@@ -558,6 +558,16 @@ export function props(
       } else if (styleProp === 'paddingInlineEnd') {
         flatStyle.paddingEnd = styleValue;
       }
+      // visibility polyfill
+      // note: we can't polyfill nested visibility changes
+      else if (styleProp === 'visibility') {
+        if (styleValue === 'hidden' || styleValue === 'collapse') {
+          flatStyle.opacity = 0;
+          nativeProps['aria-hidden'] = true;
+          nativeProps.pointerEvents = 'none';
+          nativeProps.tabIndex = -1;
+        }
+      }
       // everything else
       else {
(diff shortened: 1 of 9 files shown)


packages/react-strict-dom/src/native/stylex/index.js
```diff
@@ -558,6 +558,15 @@ export function props(
       } else if (styleProp === 'paddingInlineEnd') {
         flatStyle.paddingEnd = styleValue;
       }
+      // visibility polyfill
+      // note: we can't polyfill nested visibility changes
+      else if (styleProp === 'visibility') {
+        if (styleValue === 'hidden' || styleValue === 'collapse') {
+          flatStyle.opacity = 0;
```
nmn, line 565: Do we have something like `aria-hidden` in RN?

nmn (approved):
LGTM.

packages/react-strict-dom/src/native/stylex/index.js
necolas: Yeah, we have a polyfill for `aria-hidden` in RN itself. I'll add that prop to the polyfill

PR #36, opened 2024-02-24
chore: Update build script to generate flow and TS types:

### LINKED ISSUE #25: TypeScript support

Generate a TypeScript libdef for the npm package

nmn:
#Fixes #25

Using StyleX as a starting point, overhauling the build scripts.

- [x] Copy the `gen-flow` wrapper around `flow-api-translator`
- [x] Set up script to generate per-file Flow and TS type-defs
- [x] Update the `package.json` to point to the new output files
- [x] Delete the existing build steps

Skipped these steps as we decided to keep flat bundles for now:

- [ ] ~~Set up a Babel translation step for publishing to NPM that doesn't pre-bundle~~
- [ ] ~~Set up a Babel translation step that translates to Haste modules~~


necolas:
Nice! I kind of want to keep flat bundling for OSS

nmn:
>  I kind of want to keep flat bundling for OSS

One downside of flat bundling is that you can take advantage of tree-shaking in modern bundlers.

However, if we stop here, we can keep the flat bundles for the JS code and only keep the separate files for the type-defs.

> This isn't a modular library

If we only publish an ESM build to NPM, we can limit what we export so that all the other files are "private".

> AFAICT the motivation for this change is because flow doesn't support automatically generating lib defs

The `flow-api-translater` package being used by the script is the Flow teams solution for generating lib-defs. The other motivation is that this lets us generate TS type-defs as well. A folder of file like this is fully supported in both Flow and TS and so we don't really gain anything by bundling them into flat files.

necolas:
There's nothing to tree-shake out of this package on web, or even native really. AFAIK, tree-shaking / DCE isn't limited to what's in different files and should work if exports within a file aren't used. So we can focus on the types side of things

nmn:
@necolas That makes things simple enough. This PR is now ready. 

I simply changed the output path from `dist/dom.js` to `dist/dom/index.js` and not the separate type files work as expected. The `dist/dom/index.d.ts` and `dist/dom/index.js.flow` files shadow the actual file and that works for both TS and Flow.

I can probably make this set up work for internal sync as well.

nmn:
Update: Made a couple of changes to ensure that no two files in the package have the same name. (Except when there are `dom` and `native` counter-parts`). This was mostly true already except:
1. There was one unused file
2. There are two versions of `flattenStyle.js` that work similarly but with different types.

For now, I've renamed one of the `flattenStyle.js` to `flattenStyleXStyles.js`, but eventually, it should be deleted.

This work has been done to be compatible with Haste for internal sync.

packages/scripts/gen-types.js
```diff
@@ -0,0 +1,162 @@
...
+              line.includes('__monkey_patch__')
+            );
+            const startIndex = lines.indexOf(line);
+            const endIndex = startIndex + 4;
+
+            outputFlowContents = lines
+              .slice(0, startIndex)
+              .concat(lines.slice(endIndex))
+              .join('\n');
+          }
```
necolas, line 80: This looks like stylex-specific logic that was copy-pasted?

packages/scripts/rewrite-imports.js
```diff
@@ -0,0 +1,68 @@
...
+
+async function rewriteImportsInFolder(inputDir: string, outputDir: string) {
+  await fs.mkdir(outputDir, { recursive: true });
+  const dirents = await fs.readdir(inputDir, { withFileTypes: true });
+  for (const dirent of dirents) {
+    // We only care about the `stylex` package for now which has
+    // a flat file tree.
+    if (dirent.isDirectory()) {
+      continue;
+    }
```
necolas, line 45: StyleX-specific

packages/scripts/rewrite-imports.js
```diff
@@ -0,0 +1,68 @@
...
+      {
+        sourceMapper: ({ module }) => module.slice(module.lastIndexOf('/') + 1)
+      }
+    );
+    if (fileName === 'stylex.js') {
+      outputFile = outputFile.replace(
+        'export default (_stylex: IStyleX);\n',
+        ''
+      );
+    }
```
necolas, line 65: StyleX-specific

nmn:
Not able to reproduce this failure locally. I don't even have the file that is erroring. :(

configs/rollup.config.js
```diff
@@ -126,7 +126,10 @@ const webConfigs = [
     external: ['react', 'react-dom', '@stylexjs/stylex'],
     input: require.resolve('../packages/react-strict-dom/src/dom/index.js'),
     output: {
-      file: path.join(__dirname, '../packages/react-strict-dom/dist/dom.js'),
+      file: path.join(
+        __dirname,
+        '../packages/react-strict-dom/dist/dom/index.js'
+      ),
```
necolas, line 132: What about the www build? That is still being written to `dist/dom.www.js`; how is internal use going to work if the types are being picked up from the new flow files but they have references to the `@stylexjs/stylex` package instead of `stylex`?

necolas (commented):
Does this patch still allow importing of types from the main package, e.g., `import type { StrictHTMLElement } from 'react-strict-dom'`?

necolas:
> Not able to reproduce this failure locally. I don't even have the file that is erroring. :(

I can reproduce once this PR is rebased on main

image

nmn:
@necolas Does this look good to merge? (I know you're on a break, so let me know if I should run this by someone else while you're away)

necolas:
There's still a few things to update before that.  I'll comment over the weekend. Also left some questions above

configs/rollup.config.js
nmn: I will be transforming the imports while copying the files over during sync.

nmn:
> Does this patch still allow importing of types from the main package, e.g., import type { StrictHTMLElement } from 'react-strict-dom'?

Anything that is exported from the entry point can be imported directly. If there are types that are currently not exported from the entrypoint that are supposed to be public, we can add them.

.gitignore
```diff
@@ -4,3 +4,4 @@ build
 coverage
 dist
 node_modules
+lib
```
necolas, line 7: This doesn't seem needed

packages/react-strict-dom/src/dom/runtime.js
```diff
@@ -123,47 +123,47 @@ const textarea: StrictReactDOMPropsStyle = [
...
-  pre,
-  section,
-  select,
-  span,
-  strong,
-  sub,
-  sup,
-  textarea,
-  ul
+  a: (a: typeof a),
```
necolas, line 126: kind of a bummer that the annotations have to be so verbose

packages/scripts/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "react-strict-dom-scripts",
+  "version": "0.5.1",
```
necolas, line 3:
```suggestion
  "version": "0.0.0",
```

packages/scripts/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "react-strict-dom-scripts",
```
necolas, line 2:
```suggestion
  "private": true,
  "name": "react-strict-dom-scripts",
```

packages/scripts/package.json
```diff
@@ -0,0 +1,14 @@
...
+  "version": "0.5.1",
+  "description": "Helper scripts for stylex monorepo.",
```
necolas, line 4:
```suggestion
  "description": "Helper scripts for React Strict DOM monorepo.",
```

packages/scripts/rewrite-imports.js
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env node
+/**
```
necolas, line 2:
```suggestion

/**
```

packages/scripts/handle-flow-modules.js
```diff
@@ -0,0 +1,86 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ *
```
necolas, line 7:
```suggestion
```

packages/scripts/gen-types.js
```diff
@@ -0,0 +1,148 @@
+#!/usr/bin/env node
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ *
```
necolas, line 8:
```suggestion
```

packages/scripts/gen-types.js
```diff
@@ -0,0 +1,148 @@
+#!/usr/bin/env node
+/**
```
necolas, line 2:
```suggestion

/**
```

necolas (commented):
The paths to the build files also need to be updated in the performance workflow file.

packages/react-strict-dom/src/dom/runtime.js
nmn: I will fix this bug in `flow-api-translator`. After that, we'll be able to clean this up again.

necolas:
> The paths to the build files also need to be updated in the performance workflow file.

^ this still needs to be done.

Also need to update the paths in the build size badges in the RSD README. They will update to the correct values once I publish the release after this patch

necolas:
I pushed fixes for the path updates.

However, there's one more path in the `.flowconfig` - https://github.com/facebook/react-strict-dom/blob/main/.flowconfig#L15. Updating that to point to the new flow entry point results in flow errors which should show up in CI for that commit

nmn:
Let me figure this out.

nmn:
> Removing the mapping (or mapping to a non-existent file, as was the case before in this PR) results in no flow errors...which is a bit concerning.

This is a known issue with Flow. It causes "untyped" code rather than type errors. We might need to add some checks to cache type-coverage regressions for this. We also need to update the Flow version. I'll do that after this PR has merged.

necolas:
Thanks! Merged

Issue #15, opened 2024-02-20
(feature) <input type="radio"> element


necolas:
### Describe the feature request

Support the [`<input type="radio">` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio). This element is not currently supported on native.

PR #40, opened 2024-02-27
#15 issue: Update createStrictDOMComponent.js

dishakemble260:
Issue: 
Support the <input type="radio"> element. This element is not currently supported on native.

Changes: 
Implemented a custom RadioButton component in createStrictDOMComponent.js to mimic the behavior of radio buttons using native components.
Conditionally rendered the RadioButton component when the input type is "radio".

### DIFF
--- a/packages/react-strict-dom/src/native/modules/createStrictDOMComponent.js
+++ b/packages/react-strict-dom/src/native/modules/createStrictDOMComponent.js
@@ -153,6 +153,54 @@ function resolveTransitionProperty(property: mixed): string[] {
   return [];
 }
 
+//custom radio button component 
+const RadioButton = ({ options, selectedOption, onSelect }) => {
+  return (
+    <View>
+      {options.map((option, index) => (
+        <Pressable
+          key={index}
+          onPress={() => onSelect(option)}
+          style={({ pressed }) => ({
+            opacity: pressed ? 0.5 : 1,
+          })}
+        >
+          <View
+            style={{
+              flexDirection: 'row',
+              alignItems: 'center',
+            }}
+          >
+            <View
+              style={{
+                height: 24,
+                width: 24,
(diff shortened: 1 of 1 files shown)


packages/react-strict-dom/src/native/modules/createStrictDOMComponent.js
```diff
@@ -153,6 +153,54 @@ function resolveTransitionProperty(property: mixed): string[] {
   return [];
 }
 
+//custom radio button component 
+const RadioButton = ({ options, selectedOption, onSelect }) => {
```
nmn, line 157: This does not work like `<input type="radio">` on the web.

There is no "options" attribute. We would need to automatically unselect other radio inputs when another with same name is selected.

We would need to integrate with the nearest form.

Etc.

packages/react-strict-dom/src/native/modules/createStrictDOMComponent.js
```diff
@@ -153,6 +153,54 @@ function resolveTransitionProperty(property: mixed): string[] {
...
+  return (
+    <View>
+      {options.map((option, index) => (
+        <Pressable
+          key={index}
+          onPress={() => onSelect(option)}
+          style={({ pressed }) => ({
+            opacity: pressed ? 0.5 : 1,
+          })}
```
nmn, line 166: All styling must use StyleX so that styles can be passed in.

nmn (changes requested):
This addition seems to add functionality but none of it is consistent with the web. It introduces non-standard, native-only API which is against the goals of this project.

necolas:
Thanks but I'm going to close this because implementing `radio` type correctly (with native accessibility) is relatively complicated, and the approach in this PR is not quite right. It's not certain that we will try to polyfill this component at all

PR #42, opened 2024-02-28
Move errorMsg function to shared folder

mfrlic:
Since there were two files that were exactly the same I moved the file to shared and updated imports, also replaced console.error(“React Strict DOM: …”) with errorMsg where applicable.


necolas:
Thanks!

PR #44, opened 2024-03-05
Revamp custom property reference resolution

vincentriemer:
Revamp custom property reference resolution

Apologies in advance for the long PR.

> [!IMPORTANT]
> The last section of this description has an RFC with some points that I'd like to resolve with y'alls input before landing

## Description

This PR marks the beginning of an overhaul to how the StyleX React Native shim parses/ resolves values to make it more flexible and robust. Here I've laid the ground work of this overhaul and use it to replace the previous implementation of CSS variable resolution.

I've taken a spec-informed approach (try to match the spec as much as it makes sense for our use case) to implementing a tokenizer, parser, and object model for CSS values. To do this I heavily referenced the specs for [CSS Syntax](https://drafts.csswg.org/css-syntax-3/) (defines the tokenization and parsing algorithms) and the [CSS Typed OM](https://drafts.css-houdini.org/css-typed-om-1/) (defines the object model).

The `CSSTokenizer` and `CSSParser` are largely implementation details in this work and all parsing should be triggered at a higher level abstraction in the object model. As for the object model this first PR defines the `CSSStyleValue` base/abstract class (doesn't really do much itself) that gets implemented by `CSSUnparsedValue` and `CSSVariableReferenceValue` which are responsible for the parsing/resolution of custom property references.

[The spec](https://drafts.css-houdini.org/css-typed-om-1/#reify-tokens) goes into a lot more detail but the long and short of how this works is that whenever you're parsing a value contains a `var(` token it should always result in a `CSSUnparsedValue` instance. A `CSSUnparsedValue` is effectively a list whose elements are either a `string` or a `CSSVariableReferenceValue`. Resolving a `CSSUnparsedValue` effectively becomes a case of replacing each instance of a `CSSVariableReferenceValue` with either a string representation of what the variable was pointing to or an empty string. Once a `CSSUnparsedValue` only contains string items you can just concatenate them all and then do all your other sort of transforms. In this PR I've only implemented the branch of the parsing logic for ones that contain variable references but in the future you'd just run the `parse` method again on the resulting string from a `CSSUnparsedValue` and it should result in a concrete type since all the `var(` tokens have been removed.

All of the infrastructure introduced above is leveraged by the `resolveVariableReferences` method in `customProperties.js` which where we walk the object map to turn the `CSSUnparsedValue` into a resolved string.

> [!NOTE]
> I frankly didn't focus too much on how this performs and was instead focusing on correctness so there are probably low-hanging-opportunities to improve perf including moving some of this work to a compiler/build step much like StyleX does on the web.

## Test Plan

A lot of this work thankfully already had test coverage so a large portion of the work was ensuring that the existing tests continue to pass — though I had to make *some* adjustments to the tests as some of the expectations were to my eyes incorrect. The majority of new tests were written for the `CSSTokenizer` which gets complete coverage over the basic cases.

I added a couple tests for the `CSSParser` and `CSSUnparsedValue` largely as a mechanism to debug why other pre-existing tests were failing but once I figured it out I felt like it would be a waste to get rid of them.

I lastly added a couple additional end-to-end-ish tests (in `css-test.native.js`) that cover some of the new use cases this PR enables.

## Request for Help/Comments

1) My changes (at least locally) seemed to have broken the flow libdef generation (the script still runs successfully but its output has flow errors) and I don't have even an inkling as to why so I'd really appreciate help debugging the issue.

2) We sort of had undefined behavior previously in regards to what we should do when we try to resolve a variable which isn't defined. Since we only handled values which solely contained a single variable we could just resolve it to `null` and let our property pruning logic handle ignoring the property in its entirety — but now when a variable could be embedded in a compound value those assumptions fall apart (e.g. a `rgb(255, 255, var(--unknown))` gets resolved to `rgb(255, 255, )`). What would y'all think about changing the approach to be: if there's a variable we can't resolve in a `CSSUnparsedValue`, we **don't** promote the `CSSUnparsedValue` to a string and then ignore any properties whose value are still a `CSSUnparsedValue`.


packages/react-strict-dom/tests/typed-om/CSSParser-test.js
```diff
@@ -0,0 +1,83 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import CSSParser, { CSSFunction, CSSPreservedToken } from '../src/CSSParser';
+
+describe('CSSParser', () => {
```
necolas, line 10: Help me understand the benefits of a custom parser over using the postcss parsers that we used in stylex?

packages/react-strict-dom/tests/css-test.native.js
```diff
@@ -1066,10 +1066,10 @@ describe('properties: custom property', () => {
         'color',
         'var(--testUnfinished'
       ])
-    ).toEqual('var(--testUnfinished');
-    expect(
+    ).toEqual(undefined);
+    expect(() =>
       resolveCustomPropertyValue(customProperties, ['color', 'var(bad--input)'])
-    ).toEqual('var(bad--input)');
+    ).toThrow(new TypeError('Invalid custom property name: bad--input'));
```
necolas, line 1072: Not sure we should be throwing errors here as on web this wouldn't crash the app

packages/react-strict-dom/tests/__snapshots__/css-test.native.js.snap-native
```diff
@@ -28,14 +28,6 @@ exports[`properties: custom property css.defineVars: tokens 1`] = `
 
 exports[`properties: custom property handles undefined variables 1`] = `{}`;
 
-exports[`properties: custom property var with hover styles 1`] = `
```
necolas, line 31: Looks like this test was accidentally deleted?

packages/react-strict-dom/tests/typed-om/CSSParser-test.js
nmn: The `postcss-value-parser` is more like a tokenizer, but it should work well for parsing variables.

I also have a work-in-progress parser here: https://github.com/nmn/css-value-parser

Also, LightningCSS is a very good parser.

packages/react-strict-dom/tests/css-test.native.js
nmn: IMO, we should warn and ignore this value to match the behaviour on the web.

packages/react-strict-dom/tests/typed-om/CSSParser-test.js
vincentriemer: This started as an exploration to better understand what the tokenizer/parser were actually doing — with a focus on targeted correctness over other parts, but I have additional ideas for improvements that if we went down the path of using postcss's value parser could keep me from making those improvements as the postcss value parser is unmaintained.

Ultimately I don't care enough to keep pushing tho so if you feel strongly enough I can rewrite this to use the postcss value parser.

packages/react-strict-dom/tests/css-test.native.js
vincentriemer: Agreed — will make this change

packages/react-strict-dom/tests/css-test.native.js
```diff
@@ -1248,8 +1250,85 @@ describe('properties: custom property', () => {
...
+    });
+  });
+
+  test('css variable declaration inside a media query', () => {
+    const customProperties = {
+      example: '42px'
+    };
+    const styles = css.create({
+      test: {
+        '@media (min-width: 400px)': {
```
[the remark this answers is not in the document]
vincentriemer: I know there's a new syntax but is this previous syntax entirely unsupported? I added this test because the logic for handling this syntax was never removed from the polyfill before and I wanted to maintain functional parity in every other part other than css variable resolution.

packages/react-strict-dom/tests/__snapshots__/css-test.native.js.snap-native
vincentriemer: The test wasn't deleted, it was just changed to check the value directly instead of doing it via a snapshot

As an aside: I REALLY don't like snapshot testing when it isn't absolutely necessary — those tests (at least semantically) only verify that the testcase's output hasn't changed instead of validating that the result is actually correct.

packages/react-strict-dom/tests/typed-om/CSSParser-test.js
necolas: Not trying to push for one or the other, trying to gather information to understand the cost/benefit of adding code we maintain vs using code we don't maintain. What are some of the additional improvements you have in mind?

packages/react-strict-dom/tests/css-test.native.js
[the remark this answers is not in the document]
necolas: It does look like we're missing unit tests for the new MQ syntax in general. The implementation probably hasn't been implemented yet. Supporting the legacy syntax on native isn't essential

packages/react-strict-dom/tests/__snapshots__/css-test.native.js.snap-native
necolas: The value of the snapshots is that they capture the entire style object (or props), which has helped catch regressions that direct assertion of values missed.

packages/react-strict-dom/tests/typed-om/CSSParser-test.js
nmn: @vincentriemer Other than LightningCSS I found all other parsers for CSS values lacking, including whatever parts of the CSS Typed OM exist so far so I understand the need for a custom parser. 

In order to avoid duplication of work in the future, let's figure out a way to combine your work and the parser I've been working on.

packages/react-strict-dom/tests/__snapshots__/css-test.native.js.snap-native
vincentriemer: Okay I agree with the intent of using them but am still not convinced with how it actually works — particularly the fact that the expected results are in a different file with no clear links to the test which generated it (you can figure it out but it's not glanceable). Would there be any concerns going with [inline snapshots](https://jestjs.io/docs/snapshot-testing#inline-snapshots) instead?

packages/react-strict-dom/tests/typed-om/CSSParser-test.js
vincentriemer: @necolas The sort of thing I was thinking in terms of improvements was — because this sort of thing is so low level and we have strict flow safety — I can refactor the work to utilize tuples instead of objects for the token representation which is something I did in instyle back in the day to improve performance. Not to mention us owning our own parser lets us prioritize adding new parsing/tokenization features as we want them.

@nmn Yeah I agree w/ the avoiding of duplicating work but I've already migrated to use postcss-value-parser (pushing the changes soon) and it actually didn't require much refactoring so moving to a different parser (at least in the short term) won't be very difficult.

packages/react-strict-dom/tests/__snapshots__/css-test.native.js.snap-native
necolas: IIRC inline snapshots don't work with prettier 3. That's the error that gets printed.

In practice the tests are always run from the CLI, and can have snapshot-specific labels. Hasn't been a problem figuring out which tests are failing for StyleX or RSD so far. Furthermore, most of the tests run across both platforms so need to be external snapshots, and that makes all test snapshots consistent within the repo.

necolas (approved):
Thanks for the PR and follow ups! LGTM

Issue #46, opened 2024-03-07
(Bug) Test code failed on Windows.


ssj9685:
### Describe the issue

When I run the test in windows I get an error:
```
Error: Custom snapshot resolver functions must transform paths consistently, i.e. expects resolveTestPath(resolveSnapshotPath('consistency_check/__tests__/example. consistency_check\__tests__\example.test.js
```

### Expected behavior

All test code must pass.

### Steps to reproduce

1. Clone the repository on Windows
2. npm install
3. npm run test

### Additional comments

On Windows, I think posix need to be removed from testPathForConsistencyCheck because there is a backslash in the path.
```
/**
 * Copyright (c) Meta Platforms, Inc. and affiliates.
 *
 * This source code is licensed under the MIT license found in the
 * LICENSE file in the root directory of this source tree.
 */

const path = require('path');

/**
 * Note that jest has a bug with custom snapshot resolvers that means
 * "resolveTestPath" is incorrectly used to resolve test paths in the other
 * jest "projects" that aren't using customer resolvers. Therefore, these
 * custom resolvers are implemented in a way that allows them to function
 * like the default resolver when resolving test paths.
 */

module.exports = {
  resolveSnapshotPath: (testPath) => {
    const snapshotPath = path.join(
      path.join(path.dirname(testPath), '__snapshots__'),
      path.basename(testPath) + '.snap-dom'
    );
    return snapshotPath;
  },

  resolveTestPath: (snapshotPath) => {
    const ext = path.extname(snapshotPath);
    const testPath = path.join(
      path.dirname(snapshotPath),
      '..',
      path.basename(snapshotPath, ext)
    );
    return testPath;
  },

  testPathForConsistencyCheck: path.posix.join(
    'consistency_check',
    '__tests__',
    'example.test.js'
  )
};

```

PR #48, opened 2024-03-07
Fix custom snapshot resolver consistency

ssj9685:
I reported the issue in https://github.com/facebook/react-strict-dom/issues/46
There was an issue with tests not running in the "Windows" environment.
"posix" may be removed unless there is a specific reason for it to only be included in "testPathForConsistencyCheck".

If there is a reason, please let me know.


necolas:
Thanks!

PR #56, opened 2024-03-15
Add support for html.strike

ccpandhare:
- Added "strike" boilerplate to web and native files
- Added strike to the `tagNames` array in html-test.js so basic tests can run
- Ran tests, generated snapshots
- Attempted to add a `<strike>` example to the Examples app, but there's some sneaky stylesheet somewhere that adds "text-decoration: none" to everything by default, and I can't change it without removing it at the source (which I can't find) or hard-coding the text-decoration (which defeats the purpose(?))

### DIFF
--- a/packages/react-strict-dom/src/dom/html.js
+++ b/packages/react-strict-dom/src/dom/html.js
@@ -313,6 +313,14 @@ export const span: React$AbstractComponent<
 > = createStrict('span', defaultStyles.span);
 
 /**
+ * "strike" (inline)
+ */
+export const strike: React$AbstractComponent<
+  StrictReactDOMProps,
+  StrictHTMLElement
+> = createStrict('strike', defaultStyles.strike);
+
+/**
  * "strong" (inline)
  */
 export const strong: React$AbstractComponent<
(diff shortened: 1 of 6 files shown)


necolas:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/strike

> This element is deprecated in HTML 4 and XHTML 1, and obsoleted in the [HTML Living Standard](https://html.spec.whatwg.org/multipage/obsolete.html#strike). If semantically appropriate, i.e., if it represents deleted content, use [`<del>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/del) instead. In all other cases use [`<s>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/s).

ccpandhare:
Good point! Totally forgot about that! I have some legacy code that uses strike and the html export didn’t have it, so I just added it

ccpandhare:
It looks like we don’t have del or s here - would you recommend I add both? @necolas

necolas:
I created a new PR (from this one) that adds these elements and a few others => #57

PR #59, opened 2024-03-24
Add Linking API combined with onPress event to mimic 'a' tag behavior

### LINKED ISSUE #58: (Feature) Implement Linking API in react native to mimic 'a' tag behavior

Clicking on the 'a' tag on the native app, generates an error `<a> "href" handling is not implemented in React Native.`
Can't we use `Linking.openUrl(url: string)` function combined with `onPress` event  to mimic tag behavior?

hichem-dahi:
Clicking on the 'a' tag on the native app, generates an error `<a> "href" handling is not implemented in React Native.`

This PR adds support for navigating both links and telephone numbers.

Fixes #58

### DIFF
--- a/packages/react-strict-dom/src/native/modules/createStrictDOMComponent.js
+++ b/packages/react-strict-dom/src/native/modules/createStrictDOMComponent.js
@@ -40,7 +40,8 @@ import {
   TextInput,
   Text,
   View,
-  Pressable
+  Pressable,
+  Linking
 } from 'react-native';
 import * as stylex from '../stylex';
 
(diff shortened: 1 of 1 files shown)


necolas:
I don't think this is quite what we want as links also need to be able to use to navigate within an app and conform to patterns used on web to implement routingwith accessible links (e.g., use of preventDefault). More like Expo router

hichem-dahi:
It's true, I didn't cover all aspects of navigation. closing this PR. 
Perhaps you can cover more of what's expected in a new issue or here #58.

PR #60, opened 2024-03-24
Allow .web.jsx, .web.ts, .web.tsx to be ignored in RSD ESLint Plugin

javascripter:
# Why

At the moment, only `.web.js` is ignored for the 'valid-styles' rule but in TypeScript projects, you'd want to ignore `.web.ts` etc. files as well.

# How
Renaming `ALLOWED_FILE` to `ALLOWED_FILES` and making it an array of allowed web file extension suffixes.
`.web.jsx`, `.web.ts`, `.web.tsx` are all allowed now because styles can be defined both in pure JS/TS files and JSX counterparts.

### DIFF
--- a/packages/eslint-plugin/src/valid-styles.js
+++ b/packages/eslint-plugin/src/valid-styles.js
@@ -20,8 +20,8 @@ import type {
   SpreadElement
 } from 'estree';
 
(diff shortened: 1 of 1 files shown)


necolas:
LGTM. Thanks

PR #69, opened 2024-04-03
show warn instead of error for non supported props

TheEhsanSarshar:


TheEhsanSarshar:
I assume this is a debatable issue, so I'll close it then

PR #70, opened 2024-04-03
Remove rollup configs for internal builds

necolas:
Internally we could use patch files to rewrite the stylex import on web, and add the comment directives to exclude the files from linting/formatting, etc.

### DIFF
--- a/configs/rollup.config.js
+++ b/configs/rollup.config.js
@@ -18,30 +18,6 @@ const babelPlugin = babel({
   configFile: require.resolve('./babel.config.js')
 });
 
-function internalLicensePlugin() {
-  const header = `/**
- * @noflow
- * @nolint
- * @noformat
- * @generated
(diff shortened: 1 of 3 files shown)


nmn (approved):
FWIW, I think it's worth removing Rollup for all builds.

necolas:
The bundles are generally smaller when pre-built, as we found with React. And this also allows us to have CI accurately monitor the size of the library.

nmn:
> The bundles are generally smaller when pre-built, 

Once the end user bundles, they'd get the same benefits? But having individual files helps if you're only use parts of the package.

We could keep generating a bundle specifically for CI?

necolas:
You can't use this library as individual files. I don't see a benefit in trying to remove bundling only to keep it for CI. I think the current approach is straightforward and works

PR #72, opened 2024-04-03
Fix html.img event bug in Fabric

necolas:
React Native's "Fabric" architecture appears to have a bug and does not populate the Image 'load' event with 'source' data.


necolas:
RN issue: https://github.com/facebook/react-native/issues/43874

PR #73, opened 2024-04-03
Remove automatic camel-casing of CSS custom properties

### LINKED ISSUE #50: Remove automatic camel-casing of CSS custom property names
### Describe the issue

The original StyleX shim added camel-casing of CSS custom property names to ease internal adoption of native theme definitions that map to web theme definitions. However, custom properties are case sensitive, and we have now implemented the StyleX theming APIs. So we should remove this inconsistency and allow variable names to be case sensitive.

### Expected behavior

Custom property names are case sensitive and do not get automatically camel cased.

### Steps to reproduce

N/A

### Test case

_No response_

### Additional comments

_No response_

necolas:
Thinking we should first update internal theme configs for native so that they don't rely on this transform. Then we can merge the PR and safely sync.

Fix #50


vincentriemer (approved):
LGTM

nmn (approved):
LGTM as well!

Victor1890:
LGTM

Issue #71, opened 2024-04-03
Flow type support for data-* props


necolas:
### Describe the feature request

Flow doesn't have support for typing arbitrary `data-*` props. This requires RSD to add Meta-specific `data-*` props on an as-needed basis, and each time it is considered a "blocker" by the team migrating components to RSD. This creates incentives to either 1) patch internally generated RSD syncs with new Flow types (complicates syncs, can cause regressions), or 2) add Meta-specific props to the OSS project code (doesn't scale to other users).

cc @jbrown215 who had ideas on how to support this from the Flow side

nmn:
While we wait for the Flow team to prioritise and implement the `Template Literal Types` feature, a stop gap solution would be to create a separate file *just* for the `data-*` properties and split it out from `StrictReactDOMProps`. 

This will avoid the need for patch files when we try to maintain a list of props used internally.

We can also shadow this file with a `.d.ts` file which uses template literal types so Typescript users don't have to deal with this Flow limitation.

PR #75, opened 2024-04-04
chore: Split out data attributes to separate files

nmn:
Stop-gap solution for #71 

This is a simple refactor that splits out all definitions of `data-*` attributes to separate files. This makes it easier to maintain a separate list of supported attributes internally without the need for patch files etc.


necolas:
I reworked this PR to:

1) not increase the size of the bundles.
2) not require us to maintain 2 lists of `data-*` props (1 for Flow, 1 for JS).
3) not include any `data-*` props by default other than `data-testid`.

Please take a look and let me know if it still looks OK to you. Thanks!

necolas:
I'll revisit this to use the approach we took for `Stringish` which would allow internal overrides without having to rely on code syncs and patch files.

flow-typed/react-strict-dom.js
```diff
@@ -0,0 +1,18 @@
...
+
+// This type allows Meta (and other users) to define data-*
+// props used by their components to work around Flow's current
+// lack of support for typing arbitrary data-* props.
+declare type ReactStrictDOMDataProps = {};
+
+// This type allows Meta to internally override it with an
+// internationalization type which is a string at runtime…
+// but Flow doesn't know that.
+declare type Stringish = string;
```
nmn, line 18: This will be a problem for the generated TS types. We should use explicit imports and avoid usage of any "globals".

necolas: Yeah we can fix that later after fixing internal and syncing

PR #76, opened 2024-04-08
Remove legacy ThemeContext export

necolas:
The theme context is a legacy API that was only used on native to set theme variables. We now have a native shim for the StyleX theme APIs, and no longer require the theme context.

This is another PR that shouldn't be merged until we've migrated internal XPR/RN code onto the StyleX theme APIs.


nmn:
> This is another PR that shouldn't be merged until we've migrated internal XPR/RN code onto the StyleX theme APIs

I don't think is gonna be fully done for months FWIW.

javascripter:
This also allows react-strict-dom components to be used as React Server Components in Next.js without babel plugin (babel is buggy in Next.js App Dir so this is quite useful).

necolas:
This PR is only to remove an unused (in OSS) export for native. It shouldn't make any difference to use of Next or RSC on web today

javascripter:
@necolas 
https://github.com/javascripter/rsd-next-rsc
Here's the repro.  

I think this PR removes `contexts` export in `dom/index.js` too, not just for native, specifically this line:
https://github.com/facebook/react-strict-dom/blob/780779a65314e2b65b4ffed0c0284afaff6e8585/packages/react-strict-dom/src/dom/index.js

This `contexts` export prevents Next.js from treating it as RSC (the below error). Removing the contexts export allows it to be used as RSC (run patch-package in the repro to apply the diff and run yarn dev and now it works in RSC)

Screenshot 2024-05-08 at 0 27 52

necolas:
I updated the patch to remove ThemeContext only from the DOM exports. This should be safe to land without causing regressions inside Meta. The entry module types will differ between web and native, but the theme context can only be imported from the native entry. We can remove the native part later once Meta updates its internal use of StyleX on web to use the StyleX theming APIs.

PR #86, opened 2024-04-11
Provide structured error/warn messages

necolas:
Improve the DX of using LogBox suppression in React Native by using a common structure for error/warn messages. This allows for easier configuration to suppress either specific messages, or categories of messages, e.g., "all unsupported properties" or "all unsupported values".


nmn (approved):
Looks good! I like the idea of consistent error message strings.

Does it make sense to create additional helper methods on `LogBox` to enforce one of the patterns?

PR #90, opened 2024-04-12
feat: Export Element and utility types from react-strict-dom

### LINKED ISSUE #88: Export element types

We need to export element types like `StrictHTMLElement`

nmn:
Resolves #88 

Creates a new file called `StrictTypes` that re-exports all the element types, DOM Props and Strict Props.

The `index.js` for both `dom` and `native` are then able to re-export this one common file.

After this PR, all the various Element or other types that might be need can be imported directly from `'react-strict-dom'`.

### DIFF
--- a/packages/react-strict-dom/src/types/StrictTypes.js
+++ b/packages/react-strict-dom/src/types/StrictTypes.js
@@ -1,0 +1,19 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow strict
+ */
+
+export type { StrictElement } from './StrictElement';
+export type { StrictHTMLElement } from './StrictHTMLElement';
+export type {
+  StrictHTMLInputElement,
+  StrictHTMLOptionElement,
+  StrictHTMLSelectElement,
+  StrictHTMLTextAreaElement
+} from './StrictHTMLFormElements';
+export type { StrictHTMLImageElement } from './StrictHTMLImageElement';
+export type { StrictNode } from './StrictNode';
--- a/packages/react-strict-dom/src/dom/index.js
+++ b/packages/react-strict-dom/src/dom/index.js
@@ -12,6 +12,8 @@ import * as css from '@stylexjs/stylex';
 import { ThemeProvider } from './modules/ThemeContext';
 import { typeof ThemeProvider as TThemeProvider } from './modules/ThemeContext';
 
+export * from '../types/StrictTypes';
+
 const contexts = { ThemeProvider: ThemeProvider as TThemeProvider };
 
 export { contexts, css, html };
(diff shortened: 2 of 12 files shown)


packages/react-strict-dom/src/types/StrictTypes.js
```diff
@@ -0,0 +1,48 @@
...
+export type { StrictProps } from './StrictProps';
+export type { StrictReactDOMProps } from './StrictReactDOMProps';
+export type { StrictReactDOMAnchorProps } from './StrictReactDOMAnchorProps';
+export type { StrictReactDOMButtonProps } from './StrictReactDOMButtonProps';
+export type { StrictReactDOMImageProps } from './StrictReactDOMImageProps';
+export type { StrictReactDOMInputProps } from './StrictReactDOMInputProps';
+export type { StrictReactDOMLabelProps } from './StrictReactDOMLabelProps';
+export type { StrictReactDOMOptionProps } from './StrictReactDOMOptionProps';
+export type { StrictReactDOMSelectProps } from './StrictReactDOMSelectProps';
+export type { StrictReactDOMTextAreaProps } from './StrictReactDOMTextAreaProps';
```
necolas, line 48: I don't think we need to export these because they can be inferred from the relevant `html.*` component export

packages/react-strict-dom/src/types/StrictTypes.js
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow strict
+ */
+
+export type {
+  StrictAnimation,
+  Animatable as StrictAnimatable
```
necolas, line 12: Probably don't want to export this

packages/react-strict-dom/src/types/StrictTypes.js
```diff
@@ -0,0 +1,48 @@
...
+  StrictWheelEventTarget,
+  StrictMouseEventTarget,
+  StrictTouchEventTarget,
+  StrictEventTarget
+} from './StrictEventTarget';
+export type { StrictHTMLElement } from './StrictHTMLElement';
+export type {
+  StrictHTMLInputElement,
+  StrictHTMLOptionElement,
+  StrictHTMLCollection,
```
necolas, line 33: Probably don't want to export this

packages/react-strict-dom/src/types/StrictTypes.js
nmn: I added them as a convenience feature and because I'm not sure what the TS equivalent of `React.ElementConfig` is. But, yeah, it's not necessary.

packages/react-strict-dom/src/types/StrictTypes.js
nmn: The Element types here might actually be useful sometimes. I'm not sure if `StrictHTMLCollection` is needed though.

Generally, I erred in exporting more types than necessary to enable better types for abstractions built on top of RSD. I can remove everything other than the `*Element` types.

efoken: Yes element types are needed, I had to use them already

necolas: Talking about StrictHTMLCollection

necolas:
Thanks, LGTM!

PR #91, opened 2024-04-16
Android doesn't like the string '0' when it expects a number

tido64:
On Android, we get a RedBox if `'0'` is passed as value to props that expect a number e.g., `borderRadius`:

Screenshot_1713265966


necolas (changes requested):
Add a unit test please

necolas (approved):
Thanks

PR #93, opened 2024-04-16
chore(stylex,tests): bump stylex to 0.6.0 and fix tests

kelset:
Following up from the release of stylex 0.6.0: https://github.com/facebook/stylex/commit/87e718e6f07359a5e7cfce59da6ea89f70917bed

This PR also takes care of fixing `npm run test`. It was originally failing because `mime-types` was not a direct dependencies, and then snapshots needed to be updated.

I also tested the examples, as far as I can tell it's working:
Screenshot 2024-04-16 at 15 01 34

It throws an error in logs saying:

```sh
Warning: Failed prop type: Invalid prop `position` of value `static` supplied to `Text`, expected one of ["absolute","relative"].`
```
 so probably it was something pre-existing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
     "jest-environment-jsdom": "^29.7.0",
     "jscodeshift": "^0.15.2",
     "lint-staged": "^13.0.3",
+    "mime-types": "^2.1.35",
     "npm-run-all": "^4.1.3",
     "prettier": "^3.2.5",
     "prettier-plugin-hermes-parser": "^0.20.1",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -48,6 +48,7 @@
     "jest-environment-jsdom": "^29.7.0",
     "jscodeshift": "^0.15.2",
     "lint-staged": "^13.0.3",
+    "mime-types": "^2.1.35",
```
necolas, line 51: Why did this need to be added as a direct dependency? It was already in the package-lock file, and we don't import it anywhere in the source code.

kelset: for some reason the test suite was erroring out because of that dependency missing - I tried to clean up the codebase and reproduce but I can't, so I'll revert that commit.

necolas:
Thanks

PR #94, opened 2024-04-16
feat: add support to `textAlignVertical`

ecreeth:
### Summary

As the title says, this PR add support to `textAlignVertical` style.

Additional to it, also fix the `verticalAlign` styles not being applied to the `textarea` element


necolas:
`textAlignVertical` is not a W3C property, and we should already have a shim in place in RN for supporting `verticalAlign`. If you want to remove those changes and submit just a bug fix, I will review the PR. Thanks

ecreeth:
Thanks for letting me know. Here's the PR: https://github.com/facebook/react-strict-dom/pull/95

PR #95, opened 2024-04-16
fix(textarea): `verticalAlign` styles not being applied

ecreeth:


necolas:
Thanks. I don't know why the CI jobs didn't run on this PR. The unit tests should have failed because snapshots weren't updated. But I fixed it locally and amended the commit.

PR #98, opened 2024-04-18
fix: same 'html' types across dom and native

nmn:
The Native version of the `html` module was using RN types instead of the common `StrictHTML*` types.

This PR fixes everything to use the same common types. Now shared code across web and native can confidently use the same types for refs, and event targets.

Our types are already constrained versions of the DOM types.


necolas (approved):
Thanks LGTM!

PR #103, opened 2024-04-19
Relax the Flow types for props

necolas:
Allow `null` or `undefined` values in most cases.

Allows for code like this, where `props.role` might be undefined:

```jsx
<html.div role={props.role} />
```

This works with React DOM, because the props are any-typed.


nmn (approved):
This is a good change!

PR #105, opened 2024-04-22
Avoid non-standard textShadow in preprocessing step

necolas:
Only transform CSS textShadow into React Native textShadow styles during the style resolving pass. Minor clean up to reduce how much of this kind of work is done in style pre-processing rather than style resolving.


nmn (approved):
Looks good.

PR #106, opened 2024-04-25
Fix polyfill for unitless lineHeight

necolas:
When a unitless lineHeight is inherited, it multiples the value of the fontSize of the inheriting element. But when an em value is inherited, it multiples the value of the fontSize of the element it was originally applied to, not the one on the inheriting element.

This patch fixes the resolved values to match web's computed values, but React Native itself renders line-height differently from web, as you can see in these screenshots. Text elements have a semi-transparent red background. React Native changes the shape of the text element box when the line box size changes, and the line box of the parent text seems to grow to match the largest nested line box.

**Web**

image

**Native**

image

### DIFF
--- a/packages/react-strict-dom/src/native/stylex/index.js
+++ b/packages/react-strict-dom/src/native/stylex/index.js
@@ -220,7 +220,7 @@ function processStyle<S: { +[string]: mixed }>(style: S): S {
   const propNames = Object.keys(result);
   for (let i = 0; i < propNames.length; i++) {
     const propName = propNames[i];
-    let styleValue = result[propName];
+    const styleValue = result[propName];
 
     if (
       CSSMediaQuery.isMediaQueryString(propName) &&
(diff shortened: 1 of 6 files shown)


packages/react-strict-dom/src/native/stylex/index.js
```diff
@@ -354,6 +341,34 @@ function resolveStyle<S: { +[string]: mixed }>(
...
+    if (propName === 'lineHeight') {
+      if (
+        typeof styleValue === 'number' ||
+        (typeof styleValue === 'string' &&
+          CSSLengthUnitValue.parse(styleValue) == null)
+      ) {
+        const lineHeightValue = parseFloat(styleValue);
+        // Only convert unitless lineHeight if fontSize exists
+        if (style.fontSize != null) {
+          if (style.fontSize instanceof CSSLengthUnitValue) {
```
mellyeliu, line 356: awesome! so now we process the multiplier into a px value based on the current font size?

necolas: We were doing that before too, but earlier in the translation process. Before, it would be converted to px straight away based on whatever the fontSize was on the element. After, it will not be converted to px until it is being rendered to an element, which should allow the unitless number to be inherited properly.

PR #108, opened 2024-04-26
Adding regex handling for negative signs

mellyeliu:
## Context
Add parsing for signage in string length unit values:
- Adding parsing for (+/-) signs
- Multiplying parsed float by -1 if sign = "-"
- Updated tests and snapshots

## Testing
Modifying tests to account for positive and negative signs 

Ran `npm run jest /packages/react-strict-dom/tests/css-test.native.js`

### DIFF
--- a/packages/react-strict-dom/src/native/stylex/CSSLengthUnitValue.js
+++ b/packages/react-strict-dom/src/native/stylex/CSSLengthUnitValue.js
@@ -9,7 +9,7 @@
 
 import type { SpreadOptions } from './SpreadOptions';
 
-const LENGTH_REGEX = /^([0-9]*[.]?[0-9]+)(em|px|rem|vh|vmax|vmin|vw)$/;
+const LENGTH_REGEX = /^(-?[0-9]*[.]?[0-9]+)(em|px|rem|vh|vmax|vmin|vw)$/;
 
 type CSSLengthUnitType = 'em' | 'px' | 'rem' | 'vh' | 'vmax' | 'vmin' | 'vw';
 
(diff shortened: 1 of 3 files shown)


nmn (approved):
Looks good to me. It should handle negative values correctly now.

@necolas Will let you merge from here?

nitishmehrotra870 (approved):
looks good

packages/react-strict-dom/src/native/stylex/CSSLengthUnitValue.js
```diff
@@ -9,7 +9,7 @@
 
 import type { SpreadOptions } from './SpreadOptions';
 
-const LENGTH_REGEX = /^([0-9]*[.]?[0-9]+)(em|px|rem|vh|vmax|vmin|vw)$/;
+const LENGTH_REGEX = /^([-+]?)([0-9]*[.]?[0-9]+)(em|px|rem|vh|vmax|vmin|vw)$/;
```
necolas, line 12: I think this only needs the optional sign in the number match, so can be simplified to

```suggestion
const LENGTH_REGEX = /^(-?[0-9]*[.]?[0-9]+)(em|px|rem|vh|vmax|vmin|vw)$/;
```

packages/react-strict-dom/src/native/stylex/CSSLengthUnitValue.js
```diff
@@ -20,8 +20,11 @@ export class CSSLengthUnitValue {
     if (match == null) {
       return null;
     }
-    const [, value, unit] = match;
-    const parsedValue = parseFloat(value);
+
+    const [, sign, value, unit] = match;
+    const signMultiplier = sign === '-' ? -1 : 1;
+    const parsedValue = signMultiplier * parseFloat(value);
+
```
necolas, line 27: Shouldn't need this code anymore if using the alternative regex

necolas:
Thanks for working on the fix! Please leave merging to maintainers next time

PR #109, opened 2024-04-30
fix: use `monospace` as default font-family for `code` instead of unsupported `Courier`

ecreeth:
### Summary

This PR fixes the next error by using `monospace`.

image

### Test plan

1. Open the app on `android`
2. See the error above
3. Apply this change
4. See the applied font


nmn (approved):
I'm surprised that Courier is unsupported. `monospace` should be safer anyway.

nmn:
Related, should we support values like `system-ui` in RSD?

PR #110, opened 2024-05-01
Feat: polyfill CSS `placeContent` support on native

ecreeth:
### Summary
This PR adds the polyfill for [`placeContent`](https://developer.mozilla.org/en-US/docs/Web/CSS/place-content)

### Test plan

EDIT:  We only support single values that are common to both properties (alignContent, justifyContent)

### DIFF
--- a/packages/react-strict-dom/tests/css-test.native.js
+++ b/packages/react-strict-dom/tests/css-test.native.js
@@ -1532,3 +1532,163 @@ describe('queries: @media', () => {
     });
   });
 });
+
+/**
+ * CSS shorthand property: placeContent
+ */
+
+describe('CSS shorthand property:', () => {
+  beforeEach(() => {
+    jest.spyOn(console, 'warn');
+    console.warn.mockImplementation(() => {});
+  });
+
+  afterEach(() => {
+    console.warn.mockRestore();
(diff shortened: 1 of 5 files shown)


necolas:
Thanks. We should only support single values that map the same to both long form properties, except for first/last baseline which is only for alignItems IIRC. This is generally how stylex and React Native work for other longforms

ecreeth:
I've updated the PR. Now only single values are supported.

packages/react-strict-dom/tests/css-test.native.js
```diff
@@ -1532,3 +1532,73 @@ describe('queries: @media', () => {
...
+      case: { placeContent: 'flex-start center' }
+    });
+    expect(css.props.call(mockOptions, styles.case)).toMatchSnapshot();
+
+    expect(console.warn).toHaveBeenCalledWith(
+      expect.stringContaining(
+        'unsupported "placeContent" style property value "flex-start center"'
+      )
+    );
+  });
```
ecreeth, line 1603: Maybe we should show a different message when multiple values are used.

necolas:
Thanks! I made a few changes and merged this into main

PR #116, opened 2024-05-08
Log error for shortform CSS properties with invalid values

necolas:
Using multiple values with shortform properties is invalid.


necolas:
This should be good enough as a stop-gap, but better logging would be done in the `create` phase so it relates back to the authored styles instead of the post-processed styles that may be non-standard and RN-specific. Something for later.

PR #117, opened 2024-05-08
Fixing the strong tag font weight handling

nitishmehrotra870:
Strong tag seems to be wrongly configured to point to `bolder` fontWeight which translates to fontWeight of 900. It should be `bold` which translates to `700`

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "fmt:report": "prettier --check \"**/*.js\" --ignore-path ./configs/.prettierignore",
     "jest": "jest --config ./configs/jest.config.js",
     "jest:report": "npm run jest",
+    "jest:update": "jest --config ./configs/jest.config.js -u",
     "lint": "npm run lint:report -- --fix",
     "lint:report": "eslint packages --config ./configs/.eslintrc",
     "test": "npm run flow && npm run lint:report && npm run jest:report",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -9,6 +9,7 @@
     "fmt:report": "prettier --check \"**/*.js\" --ignore-path ./configs/.prettierignore",
     "jest": "jest --config ./configs/jest.config.js",
     "jest:report": "npm run jest",
+    "jest:update": "jest --config ./configs/jest.config.js -u",
```
necolas, line 12: `npm run jest -- -u`

https://jestjs.io/docs/cli#using-with-package-manager

packages/react-strict-dom/src/dom/runtime.js
```diff
@@ -71,7 +71,7 @@ const styles = stylex.create({
     borderStyle: 'solid'
   },
   strong: {
-    fontWeight: 'bolder'
+    fontWeight: 'bold'
```
necolas, line 74: Interesting. In the past the `strong` element was `bolder` by default in UA style sheets, as the HTML spec expects https://html.spec.whatwg.org/#phrasing-content-3

But it seems that some time over the last few years, browsers starting using `bold` - this is the case today in Chrome and Safari.

PR #121, opened 2024-05-14
feat: Export the `createStrictDOMComponent` function

nmn:
React Strict DOM today provides a set of components under the `html.` namespace that wraps the platform specific components (divs on web, View on RN).

These components add a bunch of features on top of the base component. Specifically, it adds the ability to accept stylex styles as the `style` prop. This is particularly useful on React Native, where there is no other way to use StyleX directly.

This PR exposes the `createStrictDOMComponent` function, which can be used to wrap any other component and give it the ability to accept StyleX styles, along with all the other bookkeeping needed to make the styles work consistently.

This enables wrapping custom components and SVG components and use them along RSD.


necolas:
I'll be adding explicit svg element support to RSD. I really don't think we should be exposing internals at this stage. I have some experience with this in RNfWeb. This function is also not appropriate for SVG shims either.

PR #123, opened 2024-05-16
Fix `transform` style processing for native

nitishmehrotra870:
**Issue Description:**
    - We are using Animated view for the implementation of button component internally which is shared between web and native both
    - This button component uses a `transform` css property which is written using stylex and the value of transform property is of type string. The value is something like this:
    ```
    const styles = stylex.create({
      root: {
        transform: 'scale(0.5)'
      }
    })
    ```
    - Although react-native added support for string `transform` in the [past](https://github.com/facebook/react-native/pull/34660/files), string is not yet supported for Animated views
    - Since our button component uses animated view with string `transform` value, this results in a redblock error for us.
    
 **FIX:**
    - In this PR, I am attempting to always parse `transform` property and convert it to into an array of object which works fine with react-native.
    - The change is being done in the stylex `props` function in [native](https://github.com/facebook/react-strict-dom/blob/main/packages/react-strict-dom/src/native/stylex/index.js#L464) which is responsible for shimming the style values compatible with react native

### DIFF
--- a/packages/react-strict-dom/src/native/modules/createStrictDOMComponent.js
+++ b/packages/react-strict-dom/src/native/modules/createStrictDOMComponent.js
@@ -42,6 +42,7 @@ import {
   Pressable
 } from 'react-native';
 import * as stylex from '../stylex';
+import { parseTransform } from './parseTransform';
 
 type ReactNativeProps = {
   ...StrictProps,
@@ -637,6 +638,14 @@ export function createStrictDOMComponent<T, P: StrictProps>(
         }
         if (nativeComponent === Pressable) {
           nativeComponent = AnimatedPressable;
+        }
+      } else {
+        // shim for `transform` property was done with the animated property in the if condition
+        // above, but if we have reached till here, it means that we are not yet able to shim `transform`
+        // property and going to check it here now.
+        const transformValue = styleProps.style.transform;
+        if (transformValue != null && typeof transformValue === 'string') {
+          styleProps.style.transform = parseTransform(transformValue);
         }
       }
 
--- a/packages/react-strict-dom/src/native/modules/useStyleProps.js
+++ b/packages/react-strict-dom/src/native/modules/useStyleProps.js
@@ -24,7 +24,8 @@ const passthroughProperties = [
   'transitionDelay',
   'transitionDuration',
   'transitionProperty',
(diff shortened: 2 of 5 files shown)


necolas:
React Native also supports the string syntax for transform. Did you encounter a bug with that via RSD?

nitishmehrotra870:
> We only have to use the object syntax for transitions because Animated doesn't work properly with the string syntax

Yes, the issue I encountered was mostly with Animated and transform property

> What does this mean? What's the spread method?

I am talking about the `props` method from stylex, which is being shimmed [here](https://github.com/facebook/react-strict-dom/blob/main/packages/react-strict-dom/src/native/stylex/index.js#L464)

packages/react-strict-dom/src/native/modules/createStrictDOMComponent.js
```diff
@@ -638,6 +639,14 @@ export function createStrictDOMComponent<T, P: StrictProps>(
...
           nativeComponent = AnimatedPressable;
         }
+      } else {
+        // shim for `transform` property was done with the animated property in the if condition
+        // above, but if we have reached till here, it means that we are not yet able to shim `transform`
+        // property and going to check it here now.
+        const transformValue = styleProps.style.transform;
+        if (transformValue != null && typeof transformValue === 'string') {
+          styleProps.style.transform = parseTransform(transformValue);
+        }
```
necolas, line 649: I don't understand what this comment is saying, but since transform always needs to be parsed now, let's do it in the `stylex` shim instead of here.

necolas (changes requested):

packages/react-strict-dom/src/native/modules/createStrictDOMComponent.js
nitishmehrotra870: I initially thought to keep it one place and add the `transform` shim only in `stylex`. If I remove the code from here and do everything in stylex, then I would have to also move the logic of `transitionProperty` and `transform` parsing from (useStyleTransition)[https://github.com/facebook/react-strict-dom/blob/c10e6a95385612e868b283b4e905e0ec94289f7d/packages/react-strict-dom/src/native/modules/useStyleTransition.js#L139] to stylex too. Is that okay?

PR #124, opened 2024-05-16
Fix `transform` style processing for native

necolas:
For simplicity (and to fix an internal edge case), we now translate all 'transform' values to the React Native object syntax. If a string 'transform' is ever passed to an Animated component, the animation will now work as expected.

Close #123

### DIFF
--- a/packages/react-strict-dom/tests/__snapshots__/css-test.native.js.snap-native
+++ b/packages/react-strict-dom/tests/__snapshots__/css-test.native.js.snap-native
@@ -454,7 +454,28 @@ exports[`properties: general textShadow 2`] = `
 exports[`properties: general transform: matrix 1`] = `
 {
   "style": {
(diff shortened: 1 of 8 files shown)


nitishmehrotra870 (approved):
looks good to me.

packages/react-strict-dom/tests/__snapshots__/css-test.native.js.snap-native
```diff
@@ -454,75 +454,171 @@ exports[`properties: general textShadow 2`] = `
...
+      {
+        "scaleY": 3,
+      },
+      {
+        "scaleZ": 4,
+      },
+      {
+        "translateX": 1,
+      },
+    ],
```
necolas, line 511: The loss of `translateY` here isn't technically a regression because the value is `1em` and wouldn't have worked before. (And the other transform types aren't supported on React Native). However we could look into supporting non-px units for translate at some point in the future

PR #126, opened 2024-05-20
Remove console messages from prod builds

necolas:
Wraps the log utils in `__DEV__` on native, so React Native can DCE the log messages in prod builds.

Makes sure that the benchmark build strips code wrapped in `__DEV__`, so we can see the impact of removing logs when processing unsupported props or styles that would otherwise log messages in dev.

### DIFF
--- a/packages/react-strict-dom/src/native/stylex/index.js
+++ b/packages/react-strict-dom/src/native/stylex/index.js
@@ -265,7 +265,9 @@ function processStyle<S: { +[string]: mixed }>(style: S): S {
 
       // React Native shadows on iOS cannot polyfill box-shadow
       if (propName === 'boxShadow') {
-        warnMsg('unsupported style property "boxShadow".');
+        if (__DEV__) {
+          warnMsg('unsupported style property "boxShadow".');
+        }
         delete result.boxShadow;
         continue;
       }
@@ -300,7 +302,11 @@ function processStyle<S: { +[string]: mixed }>(style: S): S {
           if (propName === 'color') {
             result['placeholderTextColor'] = styleValue[propName];
           } else {
-            warnMsg(`unsupported "::placeholder" style property "${propName}"`);
+            if (__DEV__) {
+              warnMsg(
+                `unsupported "::placeholder" style property "${propName}"`
+              );
+            }
           }
         }
         delete result['::placeholder'];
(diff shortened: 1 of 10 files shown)


necolas:
The benchmark for unsupported styles in 60% faster.

But the regression in `css.create() small style(s)` show up locally too. I can't see why that would happen. That benchmark is only for 2 style properties, and all we did was drop logging code from the build. I'll diff the builds later to see if anything else changed.

packages/react-strict-dom/src/native/stylex/index.js
```diff
@@ -300,7 +302,11 @@ function processStyle<S: { +[string]: mixed }>(style: S): S {
           if (propName === 'color') {
             result['placeholderTextColor'] = styleValue[propName];
           } else {
-            warnMsg(`unsupported "::placeholder" style property "${propName}"`);
+            if (__DEV__) {
+              warnMsg(
+                `unsupported "::placeholder" style property "${propName}"`
+              );
+            }
```
necolas, line 309: The "small" benchmark gets a little slower when this `warnMsg` or (less so) when the one for `boxShadow` is commented out...even if the `warnMsg` function is a no-op. Given this code is never reached for those benchmarks, it must just be a bizarre artifact that doesn't represent any real-world issue.

necolas (commented):
Nothing suspicious in the builds diff. I went through line by line to identify which change altered the results for the "smallStyle" create benchmark and I think it's unlikely to be a real-world issue.

packages/react-strict-dom/src/native/stylex/index.js
bgirard: I'm not profiling with a benchmark. I was profiling a real world scenario. Is it possible that my real world scenario just uses more style code that triggers this warning more often?

nmn (approved):
LGTM

packages/react-strict-dom/src/native/stylex/index.js
necolas: No. The benchmark that includes this style gets faster with the logs removed - as expected. The bizaree thing is that a different benchmark with 2 basic styles show a regression, but it never even reaches this code. As mentioned above, the regression only occurred when the log function is removed but not if the body of the log function itself is commented out - so it's unrelated to logging itself and doesn't really make sense to me.

PR #127, opened 2024-05-21
Avoid creating copies of customProperties in native polyfill

necolas:
The customProperties map can get very large, and creating copies every render can lead to avoidable GC. This diff removes most places where copies are being made.

### DIFF
--- a/packages/react-strict-dom/src/native/modules/createStrictDOMComponent.js
+++ b/packages/react-strict-dom/src/native/modules/createStrictDOMComponent.js
@@ -403,6 +403,8 @@ export function createStrictDOMComponent<T, P: StrictProps>(
        * Resolve the style props
        */
       const inheritedStyles: ?Style = React.useContext(InheritableStyleContext);
+      const inheritedCustomProperties = React.useContext(ThemeContext);
+
       const [extractedStyles, customPropertiesFromThemes] = extractStyleThemes(
         props.style
       );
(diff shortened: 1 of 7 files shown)


necolas:
I did a visual check in the examples app and the theming examples are unchanged there too.

packages/react-strict-dom/src/native/modules/createStrictDOMComponent.js
```diff
@@ -403,6 +403,8 @@ export function createStrictDOMComponent<T, P: StrictProps>(
        * Resolve the style props
        */
       const inheritedStyles: ?Style = React.useContext(InheritableStyleContext);
+      const inheritedCustomProperties = React.useContext(ThemeContext);
```
necolas, line 406: Since we already read this context here, I'm passing it to `useStyleProps` rather than reading it again from within the hook

packages/react-strict-dom/src/native/modules/createStrictDOMComponent.js
```diff
@@ -676,14 +679,16 @@ export function createStrictDOMComponent<T, P: StrictProps>(
...
-            ...inheritedCustomProperties,
-            ...customPropertiesFromThemes
-          }}
+          value={
+            customPropertiesFromThemes != null
+              ? {
+                  ...inheritedCustomProperties,
+                  ...customPropertiesFromThemes
+                }
+              : inheritedCustomProperties
```
necolas, line 690: Avoid creating a new object if we don't have any new custom properties passed in via the `style` prop for this element. In another diff we can skip the provider wrapper too, but trying to limit this change to just object copies for the purposes of assessing the hypothesis we had from profiling.

PR #129, opened 2024-05-27
Add support for default color token values from DSP

Prakshal-Jain:
The color token values have the following structure:
```tsx
   "positive": {
      "default": "#31A24C",
      "@media (prefers-color-scheme: dark)": "#31A24C"
   },
```

However, `resolveVariableReferenceValue` function only support `string` type. All other value types falls to the else statement and returned as-it-is. Currently, This leads to the `color` value to be of the type `[Object object]`, which cannot be parsed by CSS.
With this simple change, we handle the case when the `variableValue` is an object, and return the default value.

TODO: Still need to return dark-theme through media queries `@media (prefers-color-scheme: dark)`.

### DIFF
--- a/apps/examples/src/tokens.stylex.js
+++ b/apps/examples/src/tokens.stylex.js
@@ -18,5 +18,8 @@ export const tokens: VarGroup<
   }>
 > = css.defineVars({
   squareColor: 'red',
-  textColor: 'darkred'
+  textColor: {
+    default: 'darkred',
+    '@media (prefers-color-scheme: dark)': 'lightred'
+  }
 });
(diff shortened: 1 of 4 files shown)


necolas (changes requested):
Any time you write code to patch an issue please include a unit test that fails without the patch. Otherwise we can't really tell what is being patched or avoid regressions in the future

necolas:
@nmn did support for defining variables using the `light-dark` function instead of these objects land in stylex yet?

nmn:
@necolas The work on `light-dark` is still ongoing.

PR #130, opened 2024-06-04
Avoid duplicate console messages

necolas:
Reduce repeated messages in the logs so there is only one warning or error for each type of issue. In the test environment, every instance is still logged.


vincentriemer (approved):
LGTM

PR #131, opened 2024-06-04
Remove unsupported props from native

necolas:
There was some discussion about the potential cost of forwarding unsupported props to React Native versus stripping them in the RSD runtime.


nmn (approved):
LGTM!

nmn:
A similar thing would probably make sense for the stylex shim where we use the list to support `@supports`.

PR #132, opened 2024-06-04
Don't render Context Providers unless needed

necolas:
Skip rendering the existing Context Providers where possible. This patch significantly reduces how many providers are rendered by default, and provides dev-time debugging names for the providers that do get rendered.

**Before**

image

**After**

image

### DIFF
--- a/packages/react-strict-dom/src/native/modules/createStrictDOMComponent.js
+++ b/packages/react-strict-dom/src/native/modules/createStrictDOMComponent.js
@@ -21,9 +21,10 @@ import type {
 } from '../../types/react-native';
 
 import * as React from 'react';
-import { FontSizeContext } from './FontSizeContext';
-import { InheritableStyleContext } from './InheritableStyleContext';
-import { ThemeContext } from './ThemeContext';
+import { CSSCustomPropertiesContext } from './CSSCustomPropertiesContext';
+import { CSSDisplayModeInsideContext } from './CSSDisplayModeInsideContext';
+import { CSSFontSizeContext } from './CSSFontSizeContext';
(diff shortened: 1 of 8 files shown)


packages/react-strict-dom/src/native/modules/createStrictDOMComponent.js
```diff
@@ -679,31 +680,56 @@ export function createStrictDOMComponent<T, P: StrictProps>(
...
+
+      if (
+        (inheritedStyles != null &&
+          typeof inheritedStyles === 'object' &&
+          Object.keys(inheritedStyles).length > 0) ||
+        (textStyles != null &&
+          typeof textStyles === 'object' &&
+          Object.keys(textStyles).length > 0)
+      ) {
+        element = (
```
vincentriemer, line 700: I think this can be optimized further (or I could be reading this wrong):

* if there are inherited styles but no text styles wouldn't that mean you can skip the provider?
* if there are text styles but no inherited styles you could set the provider value just to textStyles, skipping the flattening
* in the case where there are both you can keep it to what it's like here already

vincentriemer (changes requested):

packages/react-strict-dom/src/native/modules/createStrictDOMComponent.js
necolas: Good point. Adding these scenarios to the unit test / app examples also surfaced that text inheritance is buggy, so I'm looking into that too.

necolas:
Closing this in favor of https://github.com/facebook/react-strict-dom/pull/133, which fixes bugs in the style inheritance polyfill by refactoring the context use

PR #133, opened 2024-06-06
Refactor style inheritance and context use

necolas:
Fixes the polyfill for style inheritance of fontSize and lineHeight. And improves performance by reducing the number of context providers rendered.

* Fix fontSize inheritance.
* Fix lineHeight inheritance if using px strings.
* Remove incorrect text style inheritance for TextInput.
* Refactor internal contexts.
* Minimize context provider rendering.

## Context tree

**Before**

image

**After**

image


packages/react-strict-dom/src/native/index.js
```diff
@@ -10,12 +10,31 @@
...
+
+function ThemeProvider(props: ProviderProps): React$MixedElement {
+  const { children, customProperties } = props;
+
+  return customProperties ? (
+    <CustomPropertiesProvider children={children} value={customProperties} />
+  ) : (
+    children
+  );
+}
```
necolas, line 33: This is just to preserve the legacy ThemeProvider export that is used internally in code written before we added StyleX theming APIs.

packages/react-strict-dom/src/native/modules/ContextInheritedStyles.js
```diff
@@ -0,0 +1,70 @@
...
+    });
+  } else {
+    computedFontSize = valueFontSize || inheritedFontSize;
+  }
+
+  const flatStyle = flattenStyle([
+    inheritedStyles as ?Style,
+    value as ?Style,
+    computedFontSize != null && { fontSize: computedFontSize }
+  ]);
```
necolas, line 61: This code now does 2 things:
1. consolidates inheritance of `fontSize` with the inheritable of other styles.
2. fixes a bug in the previous `fontSize` inheritance implementation that didn't polyfill nested `em` lengths correctly

packages/react-strict-dom/src/native/modules/createStrictDOMComponent.js
```diff
@@ -485,10 +499,11 @@ export function createStrictDOMComponent<T, P: StrictProps>(
         // Add default text styles
         nativeComponent === Text && styles.userSelectAuto,
         // Provided styles
-        nativeComponent === Text || nativeComponent === TextInput
+        nativeComponent === Text
```
necolas, line 502: `TextInput` was incorrectly inheriting text styles, but this is not how `input` behaves on web.

packages/react-strict-dom/tests/__snapshots__/html-test.native.js.snap-native
```diff
@@ -1312,57 +1313,27 @@ exports[`<html.*> style polyfills default inherited text styles 1`] = `
         {
           "color": "red",
           "fontFamily": "Arial",
-          "fontSize": 32,
+          "fontSize": 20,
           "fontStyle": "italic",
           "fontVariant": "variant",
           "fontWeight": "300",
-          "lineHeight": 48,
+          "lineHeight": 30,
```
necolas, line 1320: The test's input value was changed from `1.5` to `30px`, and before the fix to the `create` unit processing this would have rendered a value of `600` instead of `30`.

packages/react-strict-dom/tests/__snapshots__/html-test.native.js.snap-native
```diff
@@ -1312,57 +1313,27 @@ exports[`<html.*> style polyfills default inherited text styles 1`] = `
...
-          "color": "red",
-          "fontFamily": "Arial",
-          "fontSize": 16,
-          "fontStyle": "italic",
-          "fontVariant": "variant",
-          "fontWeight": "300",
-          "lineHeight": 24,
           "position": "static",
-          "textAlign": "right",
-          "textTransform": "uppercase",
```
necolas, line 1365: This is a fix to match web, where font styles set on a wrapping div don't get inherited by text inputs

packages/react-strict-dom/tests/html-test.native.js
```diff
@@ -113,46 +113,100 @@ describe('<html.*>', () => {
...
+          <html.span>text</html.span>
+          <html.span style={styles.nested}>text</html.span>
+        </html.div>
+      );
+      const rootElement = root.toJSON();
+      const firstSpan = rootElement.children[0];
+      const secondSpan = rootElement.children[1];
+      expect(firstSpan.props.style.fontSize).toBe(2 * 16);
+      // check that 'em' values are correctly inherited
+      expect(secondSpan.props.style.fontSize).toBe(1.5 * 2 * 16);
```
necolas, line 174: Before the fixes in this patch, this would have failed as the rendered value would have been `1.5 * 16`

packages/react-strict-dom/src/native/stylex/index.js
```diff
@@ -286,10 +296,7 @@ function processStyle<S: { +[string]: mixed }>(style: S): S {
 
       const maybeLengthUnitValue = CSSLengthUnitValue.parse(styleValue);
       if (maybeLengthUnitValue != null) {
-        result[propName] =
-          maybeLengthUnitValue[1] === 'px'
-            ? maybeLengthUnitValue[0]
```
necolas, line 291: Removing this fixes the unitless lineHeight polyfill. Before, it resulted in eager conversion of values like `24px` to `24`, which would later be interpreted as a unitless value even though the initial value used pixels.

packages/react-strict-dom/tests/__snapshots__/html-test.native.js.snap-native
```diff
@@ -1312,57 +1313,27 @@ exports[`<html.*> style polyfills default inherited text styles 1`] = `
         {
           "color": "red",
           "fontFamily": "Arial",
-          "fontSize": 32,
+          "fontSize": 20,
```
necolas, line 1316: This values changed only because I modified the text fixture styles

packages/react-strict-dom/src/native/modules/ContextInheritedStyles.js
```diff
@@ -0,0 +1,70 @@
...
+    useWindowDimensions();
+
+  let computedFontSize = null;
+  if (valueFontSize instanceof CSSLengthUnitValue) {
+    computedFontSize = valueFontSize.resolvePixelValue({
+      viewportHeight,
+      viewportWidth,
+      fontScale: PixelRatio.getFontScale(),
+      inheritedFontSize: inheritedFontSize
+    });
```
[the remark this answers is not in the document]
necolas: Good spot. I don't think we have tests for that and it will probably fail. I'll look into it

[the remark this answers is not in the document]
necolas: Should be resolved by https://github.com/facebook/react-strict-dom/pull/133/commits/df295e2e02c89bedfb28adcadab473692e07ef36

vincentriemer (approved):

PR #137, opened 2024-06-11
Fix text truncation ellipsis on Android

necolas:
Android does not support ellipsis overflows for text that is selectable. This patch trades off user-selection of text for the correct visual of rendering an ellipsis when text overflows the specified number of lines.

### DIFF
--- a/packages/react-strict-dom/COMPATIBILITY.md
+++ b/packages/react-strict-dom/COMPATIBILITY.md
@@ -553,7 +553,7 @@ Note these APIs can only be accessed using `Node.getRootNode().defaultView`, in 
 | justifySelf | ❌ | ❌ | |
 | left | ✅ | ✅ | |
 | letterSpacing | ✅ | ✅ | |
-| lineClamp | 🟡 | 🟡 | |
+| lineClamp | 🟡 | 🟡 | Disables text-selection on Android ([#136](https://github.com/facebook/react-strict-dom/issues/136)) |
 | lineHeight (unitless) | 🟡 | 🟡 | |
 | margin | ✅ | ✅ | |
 | marginBlock | 🟡 | 🟡 | |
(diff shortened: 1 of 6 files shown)


packages/react-strict-dom/src/native/modules/createStrictDOMComponent.js
```diff
@@ -545,6 +546,16 @@ export function createStrictDOMComponent<T, P: StrictProps>(
         style: { ..._styleProps.style }
       };
 
+      // Workaround: Android doesn't support ellipsis truncation if text is selectable
+      // See #136
+      if (
+        Platform.OS === 'android' &&
+        styleProps.numberOfLines != null &&
+        styleProps.style.userSelect === 'auto'
```
mellyeliu, line 554: can you clarify why we need to filter when `styleProps.style.userSelect === 'auto'`?

packages/react-strict-dom/COMPATIBILITY.md
```diff
@@ -553,7 +553,7 @@ Note these APIs can only be accessed using `Node.getRootNode().defaultView`, in
 | justifySelf | ❌ | ❌ | |
 | left | ✅ | ✅ | |
 | letterSpacing | ✅ | ✅ | |
-| lineClamp | 🟡 | 🟡 | |
+| lineClamp | 🟡 | 🟡 | Disables text-selection on Android ([#136](https://github.com/facebook/react-strict-dom/issues/136)) |
```
mellyeliu, line 556: thanks for adding this documentation!

packages/react-strict-dom/src/native/modules/createStrictDOMComponent.js
necolas: If `userSelect` is already `none`, we don't need to do anything.

PR #140, opened 2024-06-16
Runtime JS optimizations for native

necolas:
Integration branch for simple optimizations of loops and addition of memoization to string parsing.

### DIFF
--- a/packages/react-strict-dom/src/native/stylex/parseShadow.js
+++ b/packages/react-strict-dom/src/native/stylex/parseShadow.js
@@ -44,18 +44,18 @@ function parseValue(str: string) {
   };
 }
 
-export type ParsedShadow = {
+export type ParsedShadow = $ReadOnly<{
   inset: boolean,
   offsetX: number | string,
   offsetY: number | string,
   blurRadius: number | string,
   spreadRadius: number | string,
   color: string | null
-};
+}>;
 
-const cache = new Map<string, Array<ParsedShadow>>();
+const cache = new Map<string, $ReadOnlyArray<ParsedShadow>>();
 
-export function parseShadow(str: string): Array<ParsedShadow> {
+export function parseShadow(str: string): $ReadOnlyArray<ParsedShadow> {
   const memoizedValue = cache.get(str);
   if (memoizedValue != null) {
     return memoizedValue;
--- a/packages/react-strict-dom/src/native/stylex/parseShadow.js
+++ b/packages/react-strict-dom/src/native/stylex/parseShadow.js
@@ -53,9 +53,20 @@ export type ParsedShadow = {
   color: string | null
 };
 
+const cache = new Map<string, Array<ParsedShadow>>();
+
 export function parseShadow(str: string): Array<ParsedShadow> {
-  return str
+  const memoizedValue = cache.get(str);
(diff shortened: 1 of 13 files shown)


necolas:
## Benchmarks `main` vs `3778149c`.

(Comparative view of https://github.com/facebook/react-strict-dom/pull/140#issuecomment-2171023874)

| Suite / Test | Before | After | %Diff |
| :--- | ---: | ---: | ---: |
| **css.create** | | | |
| small | 1,001,249 | 1,066,399 | 1.06 |
| small with units | 205,737 | 497,814 | 2.42 |
| small with variables | 40,260 | 662,168 | 16.45 |
| several small | 172,170 | 344,509 | 2.00 |
| large | 257,670 | 306,734 | 1.19 |
| large with polyfills | 27,291 | 225,308 | 8.25 |
| complex | 21,391 | 120,435 | 5.63 |
| unsupported | 42,054 | 239,348 | 5.69 |
| **css.createTheme** | | | |
| simple theme | 222,406 | 221,601 | 1.0 |
| polyfill theme | 209,965 | 209,990 | 1.0 |
| **css.props** | | | |
| small | 187,018 | 186,879 | 1.0 |
| small with units | 160,623 | 162,063 | 1.0 |
| small with variables | 81,535 | 93,499 | 1.15 |
| small with variables of units | 50,191 | 69,824 | 1.39 |
| large | 62,080 | 61,807 | 1.0 |
| large with polyfills | 21,107 | 24,337 | 1.15 |
| complex | 13,791 | 14,935 | 1.08 |
| unsupported | 62,116 | 65,469 | 1.05 |
| simple merge | 150,034 | 152,422 | 1.01 |
| wide merge | 10,839 | 11,704 | 1.08 |
| deep merge | 10,816 | 11,608 | 1.07 |
| **internals** | | | |
| extractStyleThemes | 449,911 | 461,092 | 1.02 |
| flattenStyle | 566,135 | 686,129 | 1.21 |

packages/react-strict-dom/src/native/stylex/parseShadow.js
```diff
@@ -53,9 +53,20 @@ export type ParsedShadow = {
   color: string | null
 };
 
+const cache = new Map<string, Array<ParsedShadow>>();
```
nmn, line 56: Should this use a `$ReadOnlyArray`?

nmn (approved):
LGTM.

vincentriemer (approved):
Looks great!

One sort of meta piece of feedback would be that if we're caching these values we should really pull out all the stops to ensure that the values aren't ever mutated in the future. Looking at the various opportunities for that to happen:

* Tuples: I believe we can denote the elements of a tuple as readonly by specifying their variance (i.e. in CSSLengthUnitValue the ParsedValue type would be [+number, +CSSLengthUnitType])
* Arrays: (@nmn already called this out) should use $ReadOnlyArray
* CSSUnparsedValue: None of the properties are directly exposed on this class and I think the only method that would mutate the underlying data is the `set` method. I don't think it's being used anywhere already so I'd suggest removing it.

necolas:
We should probably also skip caching when `processStyle` is used to transform dynamic styles (`styles.foo(...)`)

PR #141, opened 2024-06-18
Avoid setting direction:auto

necolas:
The style 'direction:auto' is not supported in React Native and causes Fabric to log warnings.


packages/react-strict-dom/src/native/modules/createStrictDOMComponent.js
```diff
@@ -740,7 +740,7 @@ const styles = stylex.create({
     boxSizing: 'content-box'
   },
   direction: (dir: string) => ({
-    direction: dir,
+    direction: dir !== 'auto' ? dir : null,
```
javache, line 743: nit: use undefined for props that don't need to be sent to native

javache (approved):

PR #150, opened 2024-06-23
fix: Define type dependencies in file

### LINKED ISSUE #113: TypeScript Omit with StrictReactDOMProps results into empty object
### Describe the issue

When using `Omit` in TypeScript it seems to omit everything. Example:

```tsx
export interface IconButtonProps extends Omit<React.ComponentProps<typeof html.button>, "children"> {
  children: React.ReactElement<IconProps>;
}
```

This results in `IconButtonProps` type having the ONLY `children`, and some stuff from `ReactStrictDOMDataProps`, but nothing else. Every other prop, coming from `html.button` is also omitted. When removing the `Omit<...>` everything works fine.

This is maybe a bug with Flow to TypeScript conversion?

### Expected behavior

Omitting props should only omit what is wanted.

### Steps to reproduce

1. Have react-strict-dom with TypeScript set up.
2. Create IconButton.tsx containing the above example code.
3. See that the interface only contain `children`

Screenshot 2024-05-06 at 17 45 24

### Test case

_No response_

### Additional comments

_No response_

nmn:
Fixes #113 

Defines `ReactStrictDOMDataProps` and `Stringish` as normal type exports instead of expecting them to defined globally. This resolve a common error that TS users would run into.


necolas:
What's the process for overriding `ReactStrictDOMDataProps`?

nmn:
For internal usage, we can swap the `TypeShims.js` file with a custom implementation. For TS, I'm not sure.

I was thinking about this further, and it might make sense to instead define a single `.d.ts` file that defines globals which can then be overridden by users of RSD.

I wonder if someone with more TS information can chime in?

necolas:
TS ignores `data-*` props (anything with a hyphen actually), so overriding is only needed for Flow.

> we can swap the TypeShims.js file with a custom implementation

That's not going to work for OSS though.

nmn:
Let me try something else...

PR #152, opened 2024-06-25
Fix string auto-wrap hack with style inheritance of themes

necolas:
This patch fixes a source of crashes in React Native and other bugs in the style resolving logic.

1. Previously, the automatic wrapping of raw text strings with `<Text>` didn't fully resolve inherited styles containing StyleX theme variables. Passing objects as style values to React Native  was responsible for the crash.

2. `customProperties` should have values typed `mixed`, since the value can be an object returned from `css.defineVars` when a dark theme Media Query is used in the variable definition.

3. `extractStyleThemes` was only extracting variable values of string type, which excluded numeric and object value theme overrides.

4. `lineHeight` was incorrectly resolved when wrapping raw text strings in a `<Text>`, because the computed `fontSize` could end up being excluded from the `lineHeight` calculation. This may also have occurred in the general case where a numeric fontSize were used.

### DIFF
--- a/packages/react-strict-dom/src/native/modules/ContextCustomProperties.js
+++ b/packages/react-strict-dom/src/native/modules/ContextCustomProperties.js
@@ -7,13 +7,13 @@
  * @flow strict
  */
 
+import type { CustomProperties } from '../../types/styles';
+
 import * as React from 'react';
 import { __customProperties } from '../stylex';
 
-type Value = $ReadOnly<{ [key: string]: string | number }>;
-
 const defaultContext = __customProperties;
-const ContextCustomProperties: React$Context<Value> =
+const ContextCustomProperties: React$Context<CustomProperties> =
   React.createContext(defaultContext);
 
 if (__DEV__) {
@@ -22,7 +22,9 @@ if (__DEV__) {
 
 export const CustomPropertiesProvider = ContextCustomProperties.Provider;
 
-export function useCustomProperties(customPropertiesFromThemes: ?Value): Value {
+export function useCustomProperties(
+  customPropertiesFromThemes: ?CustomProperties
+): CustomProperties {
   const inheritedCustomProperties = React.useContext(ContextCustomProperties);
   if (customPropertiesFromThemes == null) {
     return inheritedCustomProperties;
--- a/packages/react-strict-dom/src/native/modules/TextString.js
+++ b/packages/react-strict-dom/src/native/modules/TextString.js
@@ -1,0 +1,41 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow strict-local
+ */
+
+import * as React from 'react';
+import { Text } from 'react-native';
+import { useCustomProperties } from './ContextCustomProperties';
+import { useInheritedStyles } from './ContextInheritedStyles';
+import { useStyleProps } from './useStyleProps';
+
+type Props = $ReadOnly<{|
+  children: string,
+  hover: boolean
+|}>;
+
+export function TextString(props: Props): React$MixedElement {
+  const { children, hover } = props;
+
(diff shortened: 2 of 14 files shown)


packages/react-strict-dom/src/native/modules/ContextCustomProperties.js
```diff
@@ -7,13 +7,13 @@
  * @flow strict
  */
 
+import type { CustomProperties } from '../../types/styles';
```
necolas, line 10: Created a new export for the the updated `CustomProperties` type. Object key values are now `mixed` since values can be objects.

packages/react-strict-dom/src/native/modules/TextString.js
```diff
@@ -0,0 +1,41 @@
...
+    customProperties,
+    hover,
+    inheritedFontSize: inheritedFontSize
+  });
+
+  return (
+    // $FlowFixMe
+    <Text {...styleProps} children={children} />
+  );
+}
```
necolas, line 41: This fixes the hack to auto-wrap text strings by making sure the inherited styles are correctly and fully resolved before being set on `<Text>`.

packages/react-strict-dom/src/native/stylex/index.js
```diff
@@ -326,10 +326,17 @@ function resolveStyle(
...
     const processedStyles = processStyle(stylesToReprocess, true);
-    Object.assign(result, resolveStyle(processedStyles, options));
+    // We can end up re-processing lineHeight without a fontSize (since it might already be resolved),
+    // which generates incorrect lineHeight values. Passing the fontSize back in avoids this.
+    const fontSizeStyle =
+      result.fontSize != null ? { fontSize: result.fontSize } : null;
+    Object.assign(
+      result,
+      resolveStyle([fontSizeStyle, processedStyles], options)
+    );
```
necolas, line 336: This is a hack to workaround a bug in how `lineHeight` is resolved. Figuring out the correct computed value for `lineHeight` involves knowing the `fontSize`, but this function can end up in a state where the `stylesToReprocess` partial contains `lineHeight` without the accompanying `fontSize` (which is in the already-computed `result` object). The quick fix (with minor perf consequence) is to forward any computed `fontSize` along with the remaining style partial when it is passed through `resolveStyle` again.

packages/react-strict-dom/tests/__snapshots__/html-test.native.js.snap-native
```diff
@@ -1554,6 +1554,38 @@ exports[`<html.*> style polyfills inherited styles 1`] = `
...
+        "position": "static",
+      }
+    }
+  >
+    <Text
+      style={
+        {
+          "color": "green",
+          "fontSize": 64,
+          "lineHeight": 192,
```
necolas, line 1579: This matches the computed styles that the fixture code generates in a web browser.

packages/react-strict-dom/tests/html-test.native.js
```diff
@@ -114,7 +114,10 @@ describe('<html.*>', () => {
         rootColor: 'green'
       });
       const nestedTheme = css.createTheme(tokens, {
-        rootColor: 'blue'
+        rootColor: {
+          default: 'blue',
+          '@media (prefers-color-scheme: dark)': 'purple'
+        }
```
necolas, line 120: Adjusting the fixture to surface regressions in handling of theme value objects. Previously this would not have set a color.

packages/react-strict-dom/tests/html-test.native.js
```diff
@@ -256,6 +259,37 @@ describe('<html.*>', () => {
...
+      });
+
+      const root = create(
+        <html.div style={[theme, styles.root]}>
+          <html.div style={styles.text}>text</html.div>
+        </html.div>
+      );
+      expect(root.toJSON()).toMatchSnapshot();
+    });
```
necolas, line 291: This is the new regression test for the React Native crash. It also regression tests the fix to inherited `lineHeight` that was found to affect auto-wrapped text.

PR #154, opened 2024-06-26
Add support for experimental React Native style implementations

necolas:
Include experimental boxShadow, filter, and mixBlendMode features. This requires React Native version detection.


vincentriemer:
So I see we have version detection but are these properties supported on both Paper and Fabric at those versions? If they are Fabric-only properties then I'd imagine we'd have to check for that too.

NickGerleman (approved):
LGTM

PR #158, opened 2024-07-03
Skip animation setup if no style transitions

necolas:
The native polyfill for CSS transitions should not setup Animated calls unless transitions are defined.


packages/react-strict-dom/tests/html-test.native.js
```diff
@@ -373,10 +376,11 @@ describe('<html.*>', () => {
       // default
       act(() => {
         root = create(<html.div />);
-        root.update(<html.div style={{}} />);
+        root.update(<html.div style={styles.none} />);
```
nmn, line 379: `null` should also work.

Issue #157, opened 2024-07-01
Update Expo SDK in example


necolas:
We use Expo to power the example / visual regression testing app. Keeping this up-to-date is important for the following reasons:

1. We validate the experience of using RSD with the latest OSS RN releases.
2. We do most RSD development in this repo, using OSS RN first.
3. We want RSD to be easy to adopt when using Expo - the recommended OSS RN framework, and a React framework that is explicitly aligned with the goals of RSD.

## The current challenges

1. We are missing developer docs for process of upgrading Expo in the RSD repo. Make it easy for anyone to improve and follow the Expo upgrade process in the RSD repo.
2. We don't have a working style extraction plugin for Metro on web. Make it so that Expo-based RSD apps can deliver static CSS files.

## Action items

* **Update to SDK 52**. Updating to SDK 52 is important for testing against the new RN architecture. The process should be documented for the future.
* **New contributor docs**. Add details to the "contributing" docs on how to maintain repo infra like Expo, Jest, etc. For Expo, I imagine it might contain links to guides like "[Upgrade Expo SDK](https://docs.expo.dev/workflow/upgrading-expo-sdk-walkthrough/)", and any other steps that may be specific to successfully upgrading the RSD repo.
* **Extract CSS in Expo**. Look into the approach taken by @javascripter's [`postcss-react-strict-dom`](https://github.com/javascripter/postcss-react-strict-dom).

necolas:
SDK report: https://github.com/expo/expo/issues/30143

EvanBacon:
> 2. Can't use latest React Native. We're stuck using whatever version of RN the SDK requires, but we're increasingly relying on features in latest releases or nighties.

Unclear what new features you're referring to here. Expo SDK 51 is the only reasonable way to access and use experimental React 19 features end-to-end on iOS, Android, and web (`npx create-expo-app -e with-canary-react-19`). 

We also test the Expo SDK against nightlies, and support new architecture. You can change or patch dependencies in dev clients however you'd like.

> 3. Installs unnecessary dependencies. We don't need RNfWeb but Expo forces this for Web.

It's unclear to me how installing `react-native-web` blocks usage with `react-strict-dom`. We enforce the installation of `react-native-web` to reduce the chance of errors on web and increase developer success. I'm more than willing to add a flag to support starting Metro web without `react-native-web` installed though.

> 4. Metro and Expo will lack StyleX integration on Web. I don't think we have a way to get the output from the StyleX CLI back into Expo.

[...]

PR #159, opened 2024-07-03
refactor: upgrade examples app to Expo SDK 51

byCedric:
This is a test to see what's required to upgrade to SDK 51, re: 

- https://github.com/facebook/react-strict-dom/issues/157
- https://github.com/expo/expo/issues/30143

So far, it seems that the multiple `react-native` versions is causing trouble. Unfortunately, this is a known issue of react-native, but definitely worth looking into. The 2nd version is installed because of two things:

- Semver starting with `0.x` considers all `0.x` versions to be breaking changes, meaning that `^0.73.y` does not include `0.74.y`.
- `react-strict-dom` has a peer dependency of `react-native: "^0.73.y`, and because its not satisfied, it auto-installs this version in the monorepo.

### DIFF
--- a/apps/examples/package.json
+++ b/apps/examples/package.json
@@ -10,14 +10,14 @@
     "dev:web": "expo start --web --clear"
   },
   "dependencies": {
-    "@expo/metro-runtime": "~3.1.3",
+    "@expo/metro-runtime": "~3.2.1",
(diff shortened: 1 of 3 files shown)


necolas:
Thanks @byCedric

PR #163, opened 2024-07-11
Fix TypeError: _version$prerelease.startsWith is not a function (it is undefined)

tomekzaw:
This PR fixes the following error that happens with `react-native@0.75.0-rc.4` and `react-strict-dom@0.0.18`.

```
TypeError: _version$prerelease.startsWith is not a function (it is undefined), js engine: hermes
```

```ts
console.log(version); // {"major": 0, "minor": 75, "patch": 0, "prerelease": 0}
```

Screenshot 2024-07-11 at 09 00 36

### DIFF
--- a/packages/react-strict-dom/src/native/stylex/isAllowedStyleKey.js
+++ b/packages/react-strict-dom/src/native/stylex/isAllowedStyleKey.js
@@ -16,7 +16,7 @@ const { major, minor, patch } = version;
 const isMain = major === 1000 && minor === 0 && patch === 0;
 // Nightly NPM package
 // $FlowFixMe (pre-release is number type)
-const isNightly = version?.prerelease?.startsWith('nightly-');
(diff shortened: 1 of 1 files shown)


necolas:
@NickGerleman More issues with the type of 'prerelease'. We changed the type to string recently but we're still setting number values in releases

NickGerleman:
iOS returning incorrect values for `version.prerelease` was fixed alongside the typings in https://github.com/facebook/react-native/commit/95d9cdf228e33d5651f41ebf5e14e80962fae118

That could be cherry-picked into RN 0.75 RC, but the issue would also go away on the final release, and next RC process.

necolas:
@tomekzaw Sounds like you can use a nightly or wait for the 0.75 release

tomekzaw:
@necolas Sure, will do, thanks!

NickGerleman:
Here is a pick request for 0.75: https://github.com/reactwg/react-native-releases/issues/371

PR #167, opened 2024-07-19
Fix inheritance of em-based lineHeight

necolas:
Make sure the computed lineHeight is inherited, unless it's a unitless value which should act as a fontSize multiplier.

### Native (before)

image

image

### Native (after)

image

image

### Web (before and after)

image

image

### DIFF
--- a/packages/react-strict-dom/src/native/modules/createStrictDOMComponent.js
+++ b/packages/react-strict-dom/src/native/modules/createStrictDOMComponent.js
@@ -40,6 +40,7 @@ import { errorMsg, warnMsg } from '../../shared/logUtils';
 import { extractStyleThemes } from './extractStyleThemes';
 import { isPropAllowed } from '../../shared/isPropAllowed';
 import { mergeRefs } from '../../shared/mergeRefs';
+import { resolveUnitlessLineHeight } from './resolveUnitlessLineHeight';
 import { useHoverHandlers } from './useHoverHandlers';
 import { useStrictDOMElement } from './useStrictDOMElement';
 import { useStyleProps } from './useStyleProps';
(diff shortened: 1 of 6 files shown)


packages/react-strict-dom/src/native/modules/createStrictDOMComponent.js
```diff
@@ -795,10 +796,8 @@ export function createStrictDOMComponent<T, P: StrictProps>(
       if (letterSpacing != null) {
         nextInheritedStyles.letterSpacing = letterSpacing;
       }
-      if (flatStyle.lineHeight != null) {
-        // Intentionally use the unresolved value to account for unitless
-        // lineHeight, which needs to be preserved when inherited.
-        nextInheritedStyles.lineHeight = flatStyle.lineHeight;
+      if (lineHeight != null) {
+        nextInheritedStyles.lineHeight = lineHeight;
```
necolas, line 800: A [previous patch](ef62857d0b8c3ce1578b03dbe581e243f21c0977) changed this logic and caused a regression when computing em-unit lineHeights that are inherited by an element setting em-unit fontSize.

packages/react-strict-dom/src/native/modules/resolveUnitlessLineHeight.js
```diff
@@ -0,0 +1,25 @@
...
+ */
+export function resolveUnitlessLineHeight(
+  style: ReactNativeStyle
+): ReactNativeStyle {
+  if (typeof style?.lineHeight === 'string') {
+    const fontSize = typeof style?.fontSize === 'number' ? style.fontSize : 16;
+    style.lineHeight = parseFloat(style.lineHeight) * fontSize;
+  }
+  return style;
+}
```
necolas, line 25: By the time we get here, the only string values left are "unitless" values that resolve with fontSize.

PR #171, opened 2024-07-20
Attempt to use Flow's new StringPrefix feature

necolas:


packages/react-strict-dom/src/types/StrictReactDOMProps.js
```diff
@@ -280,9 +280,9 @@ type EventProps = $ReadOnly<{
 }>;
 
 export type StrictReactDOMProps = $ReadOnly<{
+  [StringPrefix<'data-'>]: ?string,
```
nmn, line 283: Did you check if the generated TS types work as expected?

If so, this LGTM.

nmn (approved):

nmn:
There seems to be a Flow bug with the `StringPrefix` type still when being used in an object type as in the PR.

necolas:
This is not ready to merge, I put it up for the flow team. We've been talking about the issues with the API and it's translation to TS. They're looking into it

PR #172, opened 2024-07-25
Add 'value' to prop types for 'li' element

necolas:


packages/react-strict-dom/src/native/modules/createStrictDOMComponent.js
```diff
@@ -672,7 +672,7 @@ export function createStrictDOMComponent<T, P: StrictProps>(
         if (spellCheck != null) {
           nativeProps.spellCheck = spellCheck;
         }
-        if (value != null) {
+        if (value != null && typeof value === 'string') {
```
omarchehab98, line 675: Wondering why we added the string check here.

Does `StrictReactDOMListItemProps`'s `value` property change from number to string for this expression to pass?

omarchehab98 (approved):

packages/react-strict-dom/src/native/modules/createStrictDOMComponent.js
necolas: The `value` prop for RN TextInput is `string` and given the way we mash all the element prop types together to type this function, it required a runtime type check

PR #186, opened 2024-08-15
Switches native View to ViewNativeComponent

necolas:
Bypass the React Native "View" component for "ViewNativeComponent". RSD currently does all the intermediate transforms that View otherwise does in React Native.


necolas:
This change decreased "Render + Commit" duration times on native by ~2% in a real-world app

PR #187, opened 2024-08-16
Update flow and fix TS types generation

### LINKED ISSUE #113: TypeScript Omit with StrictReactDOMProps results into empty object

necolas:
When generating TS types: 

* Replace `Stringish` with `string`.
* Remove `ReactStrictDOMDataProps`.

Fix #113
Close #155


necolas:
cc @efoken

PR #190, opened 2024-08-22
Optimized Fast Path For re-renders of divs with unchanged contexts

rhagigi:
If I'm not missing anything, any time any RSD div re-renders for any reaason, all RSD components below it will re-render, no matter what, because this context *always* changes. This adds a fast path so that at the very least, this will now only happen when you re-render a div that actually *does* include any new inheritable styles. 

It also adds a tiny bit of a bail out (no need to object.assign to a new object) if the new inheritable styles are added to a div that has no parent inherited styles (so the parent is using the default context value, which now will get transferred down thanks to the first if). But it'll still be a new object there, because a new object will be created for "value" every time.

The way to fix this properly would be to actually do some kind of proper memoization of this value - but this hack is quick and dirty and *most* of the time won't result in re-rendering the entire subtree of RSD components, so I'll take it for now. 

Will see if I can add tests in the morning but wanted to get this out there first to take a look at as it should be fairly self explanatory.

### DIFF
--- a/packages/react-strict-dom/src/native/modules/ContextInheritedStyles.js
+++ b/packages/react-strict-dom/src/native/modules/ContextInheritedStyles.js
@@ -10,7 +10,9 @@
 import type { Style } from '../../types/styles';
 
 import * as React from 'react';
+import { useMemo, useState } from 'react';
 import { flattenStyle } from './flattenStyle';
+import { shallowEqual } from './shallowEqual';
 
 type Value = Style;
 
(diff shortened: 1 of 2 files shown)


packages/react-strict-dom/src/native/modules/ContextInheritedStyles.js
```diff
@@ -34,7 +34,17 @@ export function ProvideInheritedStyles(
...
+  if (
+    value == null ||
+    (typeof value === 'object' && Object.keys(value).length === 0)
+  ) {
+    flatStyle = inheritedStyles;
+  } else if (inheritedStyles === defaultContext) {
+    flatStyle = value;
+  } else {
+    flatStyle = flattenStyle([inheritedStyles as ?Style, value as ?Style]);
+  }
```
josephsavona, line 47: Let's make this always return the same object if the result hasn't changed. We can borrow Relay's recycleNodesInto() utility and do the following:

```js
const [flattenedStyles, setFlattenedStyles] = useState(null);
const nextFlattenedStyles = recycleNodesInto(flattenedStyles, flattenStyle(inheritedStyles, value));
if (nextFlattenedStyles !== flattenedStyles) {
  setFlattenedStyles(nextFlattenedStyles);
}
return <ContextInheritedStyles.Provider value={flattenedStyles} />
```

josephsavona: actually, it's probably also worth wrapping the `recycleNodes(..., flattenStyle(...))` bit in a useMemo just to save re-evaluating if the component happens to re-render

necolas:
When it comes to testing that this avoids re-mounts, we could add a new test below this one that checks a child is the same on both sides of a react update https://github.com/facebook/react-strict-dom/blob/main/packages/react-strict-dom/tests/html-test.native.js#L190

necolas:
Applied by #191

PR #195, opened 2024-08-27
fix: Add textDecoration properties to list of inheritable styles

nmn:
Just adds the additional styles that should be inherited.

Also fixes style flattenning logic to support legacy "@" and ":" styles in StyleX.


necolas:
Needs unit tests please.

> Also fixes style flattening logic to support legacy "@" and ":" styles in StyleX.

We don't want to do this.

necolas (changes requested):
See above

PR #198, opened 2024-08-29
Remove `experimental_` prefix from `boxShadow` and `filter`

jorge-cab:
The prefix has been fully removed from react-native so now fully removing it from RSD

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog
 
+## 0.0.27 (Sep 16, 2024)
+
+### New features
+
+* [Native] Allow use of `outline-*` styles on native (new React Native support).
+
+### Fixes
+
+* [Native] Fix support for `%` translate transforms.
+
 ## 0.0.26 (Sep 05, 2024)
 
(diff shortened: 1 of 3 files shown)


packages/react-strict-dom/src/native/stylex/index.js
```diff
@@ -632,12 +632,10 @@ export function props(
     // boxShadow
     else if (styleProp === 'boxShadow') {
       nextStyle.boxShadow = styleValue;
-      nextStyle.experimental_boxShadow = styleValue;
     }
     // filter
     else if (styleProp === 'filter') {
       nextStyle.filter = styleValue;
-      nextStyle.experimental_filter = styleValue;
     }
```
necolas, line 639: The conditions can be removed entirely once we don't need the `experimental_` prefix.

necolas (commented):
This can be merged once RN 0.76 is out with support for the unprefixed properties

NickGerleman:
> This can be merged once RN 0.76 is out with support for the unprefixed properties.

First RC for this released today

PR #203, opened 2024-09-18
Reorg tests and package entries

necolas:
* Minor reorganization of stray unit tests
* Reorganize babel code and tests
* Use Node.js package exports


nmn (approved):
LGTM!

PR #205, opened 2024-09-23
Remove _experimental prefix from boxShadow and filter

jorge-cab:
experimental_ for boxShadow and filter has been removed from React Native so now we can do the same for RSD

### DIFF
--- a/packages/react-strict-dom/src/native/stylex/index.js
+++ b/packages/react-strict-dom/src/native/stylex/index.js
@@ -629,16 +629,6 @@ export function props(
       nextStyle.alignContent = nextStyle.alignContent ?? styleValue;
       nextStyle.justifyContent = nextStyle.justifyContent ?? styleValue;
     }
-    // boxShadow
-    else if (styleProp === 'boxShadow') {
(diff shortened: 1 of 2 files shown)


packages/react-strict-dom/src/native/stylex/index.js
```diff
@@ -632,12 +632,10 @@ export function props(
     // boxShadow
     else if (styleProp === 'boxShadow') {
       nextStyle.boxShadow = styleValue;
-      nextStyle.experimental_boxShadow = styleValue;
     }
     // filter
     else if (styleProp === 'filter') {
       nextStyle.filter = styleValue;
-      nextStyle.experimental_filter = styleValue;
     }
```
necolas, line 639: These code blocks can be removed entirely

necolas (changes requested):
There's also a change to make in the function that checks for allowed style properties, as we would not longer need to conditionally only allow `boxShadow` and `filter` for nightly/main

NickGerleman:
@jorge-cab are you running into any issues with the requested changes?

PR #206, opened 2024-09-25
Prepare to remove experimental_ prefix from mix-blend-mode

jorge-cab:
We are looking into removing experimental_ from mix-blend-mode. Setting RSD to handle both prop names for now to later clean up


NickGerleman:
The `experimental` prefix was removed from this in RN. @necolas we would need to merge this for `mixBlendMode` in RSD to work in the builds which support the feature.

PR #210, opened 2024-10-03
Babel preset with integrated StyleX

necolas:
This patch changes RSD to export a babel preset (not plugin), which integrates StyleX to simplify developer setup and ensure that the correct configuration is used.

StyleX itself is also patched to fix theming on web and remove deprecated or unrelated APIs for cross-platform React.

### DIFF
--- a/packages/react-strict-dom/babel/index.js
+++ b/packages/react-strict-dom/babel/index.js
@@ -6,7 +6,8 @@
  */
 
 const path = require('path');
-const { addNamed, addNamespace } = require('@babel/helper-module-imports');
+const { addNamed } = require('@babel/helper-module-imports');
+const styleXPlugin = require('@stylexjs/babel-plugin');
 
 function createShortFilename(absolutePath, baseDir = process.cwd()) {
   if (!path.isAbsolute(baseDir)) {
@@ -18,7 +19,7 @@ function createShortFilename(absolutePath, baseDir = process.cwd()) {
   return shortPath;
 }
 
-module.exports = function ({ types: t }, options = {}) {
+function reactStrictPlugin({ types: t }, options = {}) {
   const packageName = 'react-strict-dom';
   const packageRuntime = 'react-strict-dom/runtime';
   const findImportDeclaration = (body, sourceValue) =>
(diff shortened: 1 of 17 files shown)


packages/react-strict-dom/src/dom/runtime.js
```diff
@@ -177,3 +177,5 @@ export const defaultStyles = {
   u: u as typeof u,
   ul: ul as typeof ul
 };
+
+export const resolveStyle = stylex.props;
```
necolas, line 181: This can changed to a bespoke styleq configuration for some minor runtime optimization (`disableMix:true`) and to avoid some bundle bloat from importing the StyleX runtime.

packages/react-strict-dom/babel/index.js
```diff
@@ -204,4 +212,44 @@ module.exports = function ({ types: t }, options = {}) {
...
+          dev: opts.dev,
+          importSources: [{ from: 'react-strict-dom', as: 'css' }],
+          runtimeInjection: opts.dev, // temporary until Expo/Metro can support built-time
+          styleResolution: 'property-specificity',
+          unstable_moduleResolution: {
+            type: 'commonJS',
+            rootDir: opts.rootDir
+            //themeFileExtension: '.cssvars.js',
+          },
+          useRemForFontSize: false
```
nmn, line 246: In RSD, we do not set `allowFontScaling: false` on `<Text>`. This means that all text grows and shrinks based on the system font-scaling setting. This is exactly the same behaviour as using `rem` units on the web.

So,
- either we should set this to `true`
- or, we should set `allowFontScaling: false` in RSD and rely on the usage of `rem` units for text that *should* respect font-scaling and manage that manually.

nmn (approved):
One comment aside, the rest of the PR looks good. That one comment is a pre-existing issue that we can resolve separately.

packages/react-strict-dom/babel/index.js
axyz: @necolas @nmn was this issue ever addressed? I don't see any usage of allowFontScaling in the code, so I'm assuming using rem/em polyfill will lead to double scaling? (as allowFontScaling is default to true on RN Text)

PR #215, opened 2024-10-09
Remove ref reads during render in CSS transition polyfill

### LINKED ISSUE #207: CSS transition polyfill reads ref during render
### Describe the issue

The React Compiler complains about reading a ref during render https://github.com/facebook/react-strict-dom/blob/main/packages/react-strict-dom/src/native/modules/useStyleTransition.js#L143

### Expected behavior

Follow current React best practices required by the compiler

### Steps to reproduce

1. Use React Compiler or linter

### Test case

_No response_

### Additional comments

_No response_

vincentriemer:
Fix #207

This diff updates the `useStyleTransition` hook to utilize the hook equivalent of the"getDerivedStateFromProps" pattern — which ultimately means to "set state" in render instead of writting/reading refs during render.

The only annoying part of this change is needing to do a deep comparison of the style objects since their identies are almost guaranteed to change every render. I've mitigated the performance implications of this by only comparing the style properties which are designated by `transitionProperties`. This does create some limitations in terms of how close we can get to exactly the browser's behaviro wrt CSS transitions but in my testing that limitation was already present before my changes so at least it isn't a regression.

This strategy also inherently provided some improvements like avoiding unecessary `interpolate` calls and unnecessary animations being triggered.

### DIFF
--- a/packages/react-strict-dom/src/native/modules/useStyleTransition.js
+++ b/packages/react-strict-dom/src/native/modules/useStyleTransition.js
@@ -9,19 +9,24 @@
 
 import type {
   Style as ReactNativeStyle,
-  StyleValue
+  StyleValue,
+  Transform
 } from '../../types/react-native';
 
-import { useEffect, useRef } from 'react';
-import { Animated, Easing } from 'react-native';
+import { useEffect, useRef, useState } from 'react';
 import { warnMsg } from '../../shared/logUtils';
-
-type TransitionProperties = $ReadOnlyArray<{
-  property: string,
-  value: StyleValue
-}>;
+import { Animated, Easing } from 'react-native';
 
 type AnimatedStyle = { [string]: ?StyleValue | $ReadOnlyArray<mixed> };
+
+type TransitionMetadata = $ReadOnly<{
+  delay: number,
+  duration: number,
+  timingFunction: string | null,
+  shouldUseNativeDriver: boolean
+}>;
+
+const INPUT_RANGE: $ReadOnlyArray<number> = [0, 1];
 
 function isNumber(num: mixed): num is number {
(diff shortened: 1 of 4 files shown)


packages/react-strict-dom/src/native/modules/useStyleTransition.js
```diff
@@ -92,319 +203,310 @@ export function useStyleTransition(style: ReactNativeStyle): ReactNativeStyle {
...
+      return animatedStyle;
+    } else if (typeof value === 'string') {
+      animatedStyle[property] = animatedValue.interpolate({
+        inputRange: INPUT_RANGE,
+        outputRange: [String(startValue), value]
+      });
+      return animatedStyle;
+    } else if (property === 'transform' && Array.isArray(value)) {
+      const transforms = value;
+      const refTransforms = startValue;
```
necolas, line 309: we can rename `refTransforms`/`startValue` to something like `prevTransforms`/`prevValue` now that it's state

nmn:
> which ultimately means to "set state" in render instead of writting/reading refs during render.

Would `useDeferredValue` work as well?

vincentriemer:
> > which ultimately means to "set state" in render instead of writting/reading refs during render.
> 
> Would `useDeferredValue` work as well?

I'm admittedly less familiar with `useDeferredValue` but after a quick read of the docs I think it falls short in a couple ways:

1) It doesn't have the ability to provide a custom equality check which this solution requires otherwise it causes render loops
2) AFAIK `useDeferredValue` actually causes multiple full tree re-renders/commits whereas this solution causes just the component which set state in render to re-render and results in only one commit.

necolas (approved):
Thanks! LGTM

PR #219, opened 2024-10-14
feat: mark `inset*`, `marginBlock*`, `marginInline*`, `paddingBlock*` and `paddingInline*` props as built-in support

ecreeth:
This layout props have been implemented in main (default as of 0.76).

Check [46478](https://github.com/facebook/react-native/pull/46478) for more information.

Should we continue to test these props?


ecreeth:
I'm getting this coverage error: 
```bash
Jest: "global" coverage threshold for branches (89%) not met: 88.79%
```

necolas:
We haven't been using the built in support because all logical styles aren't fully implemented in RN yet, and we had problems in the past with missing desktop support. I'll ask around and see if we can start to drop some of the polyfills

packages/react-strict-dom/tests/__snapshots__/css-test.native.js.snap-native
```diff
@@ -1011,6 +1008,7 @@ exports[`properties: logical direction inset: inset vs top 1`] = `
 {
   "style": {
     "bottom": 100,
+    "insetBlockStart": 3,
```
necolas, line 1011: This looks like a regression

packages/react-strict-dom/tests/__snapshots__/css-test.native.js.snap-native
```diff
@@ -1031,6 +1028,7 @@ exports[`properties: logical direction inset: insetBlock vs top 1`] = `
...
```
necolas, line 1031: Here too

packages/react-strict-dom/tests/__snapshots__/css-test.native.js.snap-native
nmn: This is how RN works, so this might be expected.

ecreeth:
Looks like a miss the `direction` prop. Now is supported on both platforms. 

See  https://github.com/facebook/react-native/blob/main/packages/react-native/Libraries/StyleSheet/StyleSheetTypes.d.ts#L114

PR #220, opened 2024-10-14
Replace `React.AbstractComponent` with component type

SamChou19815:
In order to adopt react 19's ref-as-prop model, we need to eliminate all the places where they are treated differently. `React.AbstractComponent` is the worst example of this, and we need to eliminate it.

This PR replaces all of them in react-strict-dom's html.js file with the new [component types](https://flow.org/en/docs/react/component-types/).

### DIFF
--- a/packages/react-strict-dom/src/dom/html.js
+++ b/packages/react-strict-dom/src/dom/html.js
@@ -25,236 +25,270 @@ import { defaultStyles } from './runtime';
 /**
  * "a" (inline)
  */
-export const a: React$AbstractComponent<
-  StrictReactDOMAnchorProps,
-  HTMLAnchorElement
-> = createStrict('a', defaultStyles.a);
+export const a: component(
+  ref?: React.RefSetter<HTMLAnchorElement>,
+  ...StrictReactDOMAnchorProps
+) = createStrict('a', defaultStyles.a);
 
 /**
  * "article" (block)
  */
-export const article: React$AbstractComponent<
-  StrictReactDOMProps,
-  HTMLElement
-> = createStrict('article', defaultStyles.article);
+export const article: component(
+  ref?: React.RefSetter<HTMLElement>,
+  ...StrictReactDOMProps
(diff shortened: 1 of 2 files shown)


nmn (approved):
Changes look good. Need to verify that the equivalent TS types are still valid.

necolas:
(BTW please leave merges to maintainers)

The TS types before:

```js
export declare const a: React.ComponentType<
  StrictReactDOMAnchorProps & React.RefAttributes<HTMLAnchorElement>
>;
```

After:

```js
export declare const a: (
  props: Omit<
    StrictReactDOMAnchorProps,
    keyof ({ ref?: React.Ref<HTMLAnchorElement> })
  > & { ref?: React.Ref<HTMLAnchorElement> }
) => React.ReactNode;
```

nmn:
This type:

```tsx
export declare const a: (
  props: Omit<
    StrictReactDOMAnchorProps,
    keyof ({ ref?: React.Ref<HTMLAnchorElement> })
  > & { ref?: React.Ref<HTMLAnchorElement> }
) => React.ReactNode;
```

... could be simplified to:

```tsx
export declare const a: (
  props: Omit<
    StrictReactDOMAnchorProps,
    'ref',
  > & { ref?: React.Ref<HTMLAnchorElement> }
) => React.ReactNode;
```

And since `StrictReactDOMAnchorProps` doesn't have a `ref` key in the type defs, this could be further simplified, to just:

```tsx
export declare const a: (
  props: 
    StrictReactDOMAnchorProps &
    { ref?: React.Ref<HTMLAnchorElement> }
) => React.ReactNode;
```

This should be correct, as long as the React TS types allow such functions to be components.
If that is not the case, the TS types would need to be updated for React 19.


----

We should be able to update `flow-api-translator` to automatically simplify:

```tsx
keyof ({ ref?: React.Ref<HTMLAnchorElement> })
```

to

```tsx
'ref'
```

PR #222, opened 2024-10-16
Replace `React.AbstractComponent` with component type from all public signatures

SamChou19815:
In order to adopt react 19's ref-as-prop model, we need to eliminate all the places where they are treated differently. `React.AbstractComponent` is the worst example of this, and we need to eliminate it.

This PR replaces all of them in react-strict-dom's public signatures with the new [component types](https://flow.org/en/docs/react/component-types/).


necolas (approved):
LGTM

PR #224, opened 2024-10-16
Replace React utility types with $ with the namespaced ones

SamChou19815:
One day, Flow will delete `React$MixedElement`, `React$Node` and all other dollar types from the global libdef, since they are considered to be internal. All these react utility types can already be used without importing React so there is no practical benefit of keeping these dollar types either.


necolas:
Thanks Sam

PR #225, opened 2024-10-22
chore: mark `borderRadius % values` as supported

ecreeth:
See https://github.com/facebook/react-native/pull/44408 and https://github.com/facebook/react-native-website/issues/4264 for more info.

> 0.76 supports border radius % on new arch.


necolas:
Thanks. We need to wait for 0.76, which I can't install in the examples app while using the current version of Expo

PR #227, opened 2024-10-24
feat: add `eslint-plugin-react-compiler`

### LINKED ISSUE #216: Add the React Compiler eslint plugin

Adding the plugin can help us work towards ensuring that the source code can be optimized by the React Compiler https://react.dev/learn/react-compiler#installing-eslint-plugin-react-compiler

ecreeth:
This PR fixes #216

Currently, we're doing a lot of `nativeProps.PROP = VALUE;` assignments and the compiler doesn't like at all. I throws:

```txt
Mutating a value returned from a function whose return value should not be mutated react-compiler/react-compiler
```

I fixed it by doing shallow copies, but I don't know if this is the right approach. If this is ok I can continue doing the rest.

Tried to dig dive a little and found this: https://github.com/facebook/react/issues/29093#issuecomment-2116484361

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -18,15 +18,13 @@
     "requireConfigFile": false,
     "sourceType": "module"
   },
-  "extends": [
-    "plugin:ft-flow/recommended",
-    "prettier"
-  ],
+  "extends": ["plugin:ft-flow/recommended", "prettier"],
   "plugins": [
     "ft-flow",
     "promise",
     "react",
-    "react-hooks"
+    "react-hooks",
+    "eslint-plugin-react-compiler"
   ],
   "env": {
     "browser": true,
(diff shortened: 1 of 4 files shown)


.eslintrc
```diff
@@ -166,6 +164,9 @@
 
     // react-hooks
     "react-hooks/rules-of-hooks": "error",
-    "react-hooks/exhaustive-deps": "error"
+    "react-hooks/exhaustive-deps": "error",
+
+    // react-compiler
+    "react-compiler/react-compiler": "warn"
```
ecreeth, line 170: `warn` until all fixes has been applied.

ecreeth:
**The benchmarks/{size, perf} jobs are failing because of this:** 
"_In public repositories this action does not work in pull_request workflows when triggered by forks_"

See more in https://github.com/peter-evans/create-or-update-comment/issues/76#issuecomment-799847814

PR #230, opened 2024-10-31
feat: add `mark` element

### LINKED ISSUE #229: (feature) <mark> element

Support the [`<mark>` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/mark).

ecreeth:
This PR fixes #229

### DIFF
--- a/packages/react-strict-dom/src/native/html.js
+++ b/packages/react-strict-dom/src/native/html.js
@@ -71,6 +71,10 @@ const styles = stylex.create({
   },
   underline: {
     textDecorationLine: 'underline'
+  },
+  mark: {
+    backgroundColor: 'yellow',
+    color: 'black'
   }
 });
 
@@ -318,6 +322,14 @@ export const main: component(
   ref?: React.RefSetter<HTMLElement>,
   ...StrictReactDOMProps
 ) = createStrict('main');
+
+/**
+ * "mark" (inline)
+ */
+export const mark: component(
+  ref?: React.RefSetter<HTMLElement>,
+  ...StrictReactDOMProps
+) = createStrictText('mark', { style: styles.mark });
 
 /**
  * "nav" (block)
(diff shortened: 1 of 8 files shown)


packages/react-strict-dom/src/native/html.js
```diff
@@ -71,6 +71,10 @@ const styles = stylex.create({
   },
   underline: {
     textDecorationLine: 'underline'
+  },
+  mark: {
+    backgroundColor: 'yellow',
+    color: 'black'
   }
```
ecreeth, line 78: Taken from https://chromium.googlesource.com/external/Webkit/+/master/Source/WebCore/css/html.css#1165

nmn:
This looks right. I verified that the default styles for a `<mark>` element don't change with with `color-scheme: dark` applied and so the styling is consistent with what is expected on the web.

Based on my testing in Expo Snack, the rendered results are as expected in Android, but slightly wrong on iOS.


Android:
Screenshot 2024-10-31 at 12 52 57 am

iOS:

Screenshot 2024-10-31 at 12 51 52 am

---

Is this a bug we should report to RN? Note how the yellow background continues to end of the line even when the text has already wrapped to the next line?

ecreeth:
> Is this a bug we should report to RN? Note how the yellow background continues to end of the line even when the text has already wrapped to the next line?

I'll check later on this locally in RN 0.76.1 and RN 0.75.* to verify if this is a regression. I'll share screenshots!

necolas:
Pretty sure that's a known RN bug

ecreeth:
I could reproduce the issue locally and confirm it only happens on `IOS`, although I only tested it on the current version of the demo app. It is indeed a RN bug.
| Android | IOS |
| -------- | ------- |
|Screenshot_1730419884 | Simulator Screenshot - iPhone 15 Pro - 2024-10-31 at 20 11 16|

PR #231, opened 2024-11-05
feat: expose the useCustomProperties hook

nmn:
When trying to style "native" (not RSD) components with StyleX APIs, it should be possible to resolve the themes that have been applied with `stylex.createTheme` APIs within the RSD tree.

This PR simply exposes the `useCustomProperties` hook to make this possible.


necolas:
This API is private implementation detail and shouldn't be exposed. We don't want arbitrary parts of the internals exposed, especially when some of these features will be built into native one day. I don't think the use case of trying to use stylex directly with RN is significant enough to introduce more problems like this. If anything, we should be doing the opposite internally and moving stylex into implementation details by using RSD exports on web

PR #232, opened 2024-11-07
Add 'compat' API to wrap RN elements

necolas:
Although it's not recommended, we have internal component forks that need to delegate to RN elements within RSD trees. This API makes that possible, while preserving the props conversion and rendering the necessary contexts to allow switching back to RSD elements within the subtree.

As implemented, it currently requires a type annotation to be used on the `nativeProps` arg.

```js
// Foo.native.js
import { compat } from 'react-strict-dom'
import { View } from 'react-native'

function Foo(props) {
  return (
    <compat.native aria-label={props.label} style={styles.foo}>
      {(nativeProps: React.PropsOf<View>) => <View {...nativeProps}>{/**/}</View>}
    </compat.native>
  )
}
```


apps/website/docs/api/04-other/02-compat.md
```diff
@@ -0,0 +1,38 @@
+# compat
+
+<p className="text-xl">Unstable components for mixing React Native with React Strict DOM elements.</p>
+
+:::warning
+
+This is an experimental and unstable API intended to aide with incremental adoption in apps that cannot avoid nesting React Native element trees within React Strict DOM trees.
```
nmn, line 7:
```suggestion
This is an experimental and unstable API intended to aid with incremental adoption in apps that cannot avoid nesting React Native element trees within React Strict DOM trees.
```

packages/react-strict-dom/src/native/modules/createStrictDOMComponent.js
```diff
@@ -142,12 +142,18 @@ export function createStrictDOMComponent<T, P: StrictProps>(
...
+          // $FlowFixMe
+          <NativeComponent {...nativeProps} />
+        );
 
       if (
-        nativeProps.children != null &&
-        typeof nativeProps.children !== 'string'
+        (nativeProps.children != null &&
+          typeof nativeProps.children !== 'string') ||
+        typeof props.children === 'function'
```
nmn, line 156: redundant?

packages/react-strict-dom/src/native/modules/createStrictDOMComponent.js
```diff
@@ -142,12 +142,18 @@ export function createStrictDOMComponent<T, P: StrictProps>(
         nativeProps.experimental_layoutConformance = 'strict';
       }
 
-      // $FlowFixMe
-      let element = <NativeComponent {...nativeProps} />;
+      let element: React.Node =
+        typeof props.children === 'function' ? (
```
nmn, line 146: So technically, any component accepts function as children now and we're using types to ensure it isn't abused this way?

nmn (approved):
Nice!

packages/react-strict-dom/src/native/modules/createStrictDOMComponent.js
necolas: useNativeProps doesn't return `children` (it made the types simpler), so the check if done against the original props

packages/react-strict-dom/src/native/modules/createStrictDOMComponent.js
necolas: That's right, but only the native components accept function children. Maybe I can add another arg to the factory functions to prevent runtime use without types

packages/react-strict-dom/src/native/modules/createStrictDOMComponent.js
nmn: I missed the `nativeProps` vs `props`. Sorry.

packages/react-strict-dom/src/native/modules/createStrictDOMComponent.js
nmn: I think this is fine, was just making sure I understood the implementation.

PR #235, opened 2024-11-12
Disentangle platform-specific flow types

necolas:
Separate the platform-specific types into `*.native` and `*.web` files. This helps generate libdefs that type check correctly in different environments.


nmn (approved):
Looks good, works without any type errors.

Issue #236, opened 2024-11-12
Update Expo SDK in example 


necolas:
# Update the Expo SDK

We use Expo to power the example / visual regression testing app. Keeping this up-to-date is important for the following reasons:

1. We validate the experience of using RSD with the latest OSS RN releases.
2. We do most RSD development in this repo, using OSS RN first.
3. We want RSD to be easy to adopt when using Expo - the recommended OSS RN framework, and a React framework that is explicitly aligned with the goals of RSD.

## The current challenges

1. We are missing developer docs for process of upgrading Expo in the RSD repo. Make it easy for anyone to improve and follow the Expo upgrade process in the RSD repo.
2. We don't have a working style extraction plugin for Metro on web. Make it so that Expo-based RSD apps can deliver static CSS files.

## Action items

- [x] **Update to SDK 52**. Updating to SDK 52 is important for testing against the new RN architecture. The process should be documented for the future. Blocks [#219](https://github.com/facebook/react-strict-dom/pull/219) and [#225](https://github.com/facebook/react-strict-dom/pull/225).
- [ ] **New contributor docs**. Add details to the "contributing" docs on how to maintain repo infra like Expo, Jest, etc. For Expo, I imagine it might contain links to guides like "[Upgrade Expo SDK](https://docs.expo.dev/workflow/upgrading-expo-sdk-walkthrough/)", and any other steps that may be specific to successfully upgrading the RSD repo.
- [x] **Extract CSS in Expo**. Look into the approach taken by @javascripter's [`postcss-react-strict-dom`](https://github.com/javascripter/postcss-react-strict-dom).

cc @ecreeth @javascripter

javascripter:
I'm happy to help advance Expo CSS extraction sitaution!
Regarding the approach used by my  plugin, here's the quick summary and the thinking behind  the plugin:

### Overview:
- Detection: Scans imported CSS files to locate the @stylex at-rule directive,
  which acts as a placeholder for generated styles (expects only one such file).
- Compilation: Transforms source files matching a specified glob pattern (from plugin options) using a Babel preset for RSD, collecting styles during this process.
- Replacement: Replaces the @stylex directive in the CSS with the compiled CSS

### Watch Mode:

1. Dependency Tracking: Watches entire directories that may contain files needing transformation, not just those matching the glob pattern. This broader watching ensures reliable detection of added or deleted files during development (PostCSS CLI can't reliably detect new files added in watch mode)
[...]

PR #238, opened 2024-11-14
Update Expo example app to enable static CSS extraction

javascripter:
# Why
ref. https://github.com/facebook/react-strict-dom/issues/236

This PR adds the ability to extract CSS in the example Expo app on the web by using [postcss-react-strict-dom](https://github.com/javascripter/postcss-react-strict-dom), as suggested in [this comment](https://github.com/facebook/react-strict-dom/issues/236#issuecomment-2475442946).

# How
1. Configure psotcss-react-strict-dom
2. Disable `runtimeInjection: opts.dev` hack and always set `runtimeInjection` to false
  - We can pass dev: false in the `metro.config.js` instead to disable runtimeInjection but that disables debug classsName generation during development, so I changed the default to false to match the output with the original branch in dev
3. Add `@expo/metro-runtime` to enable Fast Refresh
  - This is required when importing global css files in Expo Web. Without this css won't update on code changes

# Known issues

### Code and Pre elements render discrepancy

`code` and `pre` elements seem smaller than the expected output. The expected font family for these elements are `monospace, monospace` ([this code](https://github.com/facebook/react-strict-dom/blob/98afb90e1c59fdb45b80b53566865c680a564106/packages/react-strict-dom/src/dom/runtime.js#L45)).

However, in my branch the applied font-family is `monospace`. `monospace, monospace` seems to be a hack to force browsers to render monospace fonts in the expected font-size but because of the property value normalization the hack gets removed.

This seems to be an issue with Expo Web CSS post-processing, as removing postcss config and just writing `html { font-family: monospace, monospace; } ` in the global css results in the same behavior.
I suspect this to be related to the usage of lightningcss in Expo but I have not confirmed it.
https://github.com/expo/expo/blob/cd39671d70ad59277af77f156920002a148ca4a3/packages/%40expo/metro-config/src/transform-worker/transform-worker.ts#L222

Actual:
Screenshot 2024-11-14 at 19 50 03

Expected:
Screenshot 2024-11-14 at 19 51 25

# Test Plan
- [x] Compare the output of the example page with the main branch by running `npm run dev:web --clear` command in the `apps/examples` directory.
  - [ ] the rendered result seem to be identical except for one issue described in Known Issues
- [x] Edit App.tsx to check that Fast Refresh still works
- [x] Run `npx expo export -p web` to check that production builds successfully run and the generated static css is outputted to `apps/examples/dist/_expo/static/css/*.css`

### DIFF
--- a/apps/examples/postcss.config.js
+++ b/apps/examples/postcss.config.js
@@ -1,0 +1,11 @@
+module.exports = {
+  plugins: {
+    'postcss-react-strict-dom': {
+      include: [
+        'src/**/*.{js,jsx,ts,tsx}',
+        '../../packages/react-strict-dom/dist/**/*.{js,jsx,ts,tsx}'
+      ]
+    },
+    autoprefixer: {}
+  }
+};
--- a/apps/examples/package.json
+++ b/apps/examples/package.json
@@ -20,6 +20,8 @@
     "react-native-web": "~0.19.13"
   },
   "devDependencies": {
-    "@babel/core": "^7.25.2"
+    "@babel/core": "^7.25.2",
+    "autoprefixer": "^10.4.20",
+    "postcss-react-strict-dom": "0.0.2"
   }
 }
--- a/apps/examples/src/global.css
+++ b/apps/examples/src/global.css
@@ -1,0 +1,2 @@
+/* This placeholder file is required for postcss-react-strict-dom */
+@stylex;
--- a/apps/examples/src/index.js
+++ b/apps/examples/src/index.js
@@ -5,11 +5,14 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+import '@expo/metro-runtime'; // Required for Fast Refresh to work on Expo Web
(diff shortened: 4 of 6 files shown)


apps/examples/postcss.config.js
```diff
@@ -0,0 +1,11 @@
+module.exports = {
+  plugins: {
+    'postcss-react-strict-dom': {
+      include: [
+        'src/**/*.{js,jsx,ts,tsx}',
+        '../../packages/react-strict-dom/dist/**/*.{js,jsx,ts,tsx}'
```
necolas, line 6: I used `require.resolve('react-strict-dom')` in #239.
See https://github.com/javascripter/postcss-react-strict-dom/issues/3

necolas:
I fixed the monospace regression by changing the value to `monospace, "monospace"`, which prevents the tooling from de-duping it.

necolas:
I added you as a co-author to #239, as I picked up a few of the edits you made in this patch.

PR #239, opened 2024-11-14
Use postcss to extract styles in Expo example

### LINKED ISSUE #34: How to extract StyleX CSS for Expo web?

The RSD example app uses Expo to target Android, iOS, and Web but does not support extracting styles to an external CSS file for web. For the purposes of the demo, we use Babel to process the styles on web but currently have to leave the runtime injection functions in the bundle.

The next step for improving Expo Web performance with StyleX would be to have a proper Metro integration that automatically extracts CSS and inserts the relevant `<link>` tag in the app shell. This could also help Expo become one of the easiest ways for React developers to replicate Meta's production setup on web, as other frameworks currently require developers to manually integrate and configure the relevant StyleX bundler plugin rather than working out-of-the-box.

(Opinionated authoring experiences, like Tailwind syntax for web and native, could then also be built on top of StyleX to avoid the need for Tailwind tooling and provide the robust style-merging mechanics of StyleX)

cc @evanbacon @nmn

necolas:
Extract styles to static CSS using postcss, as there is no existing Metro integration.

Fix #34
Ref #236


necolas:
Thanks @javascripter for the quick patches! LGTM now. Let me know what you think

javascripter:
@necolas Thanks for testing out. Looks good to me as well!

PR #249, opened 2024-12-22
Add missing 'dist' folder before publish

Yusang-park:
This exports the dist folder, but it does not generate the dist folder during publish, resulting in the error: 
```shell
Cannot find module 'react-strict-dom' or its corresponding type declarations.ts(2307).
``` 
To fix this issue, I added a `prepublishOnly` script to perform the build process before publishing.

### DIFF
--- a/packages/react-strict-dom/package.json
+++ b/packages/react-strict-dom/package.json
@@ -18,7 +18,8 @@
     "dev": "npm run build -- --watch",
     "jest": "jest --config ./tools/jest.config.js",
     "jest:report": "jest --config ./tools/jest.config.js --collect-coverage",
-    "prebuild": "npm run clean && generate-types -i src/ -o dist"
+    "prebuild": "npm run clean && generate-types -i src/ -o dist",
+    "prepublishOnly": "npm run build"
   },
   "dependencies": {
     "@stylexjs/stylex": "^0.9.3",


necolas:
`npm install` has to be run before any publish, which runs `npm run build`

javascripter:
I'm seeing an error when upgrading to v0.0.30.
v0.0.30 is missing `dist` dir from the published artifacts despite package.json exports refer to the `dist/` dir. I think the error the author refers to is about the current published version.
Screenshot 2024-12-24 at 0 47 29

In v 0.0.29, `dist/` exists.
Screenshot 2024-12-24 at 0 49 19

Yusang-park:
@necolas 
As @javascripter mentioned, installing the latest version of the package via npm install shows that the dist folder does not exist within the package.

I’m not sure if the dist was manually built before publishing up until now, but adding this script to automatically generate the dist folder could be helpful.

necolas:
> v0.0.30 is missing dist dir from the published artifacts

Thanks for point this out. It would have been better to open an issue about this instead of submitting an unclear PR.

I tried publishing 0.0.31, checking that `dist` was there before (it's always in the file tree on my local machine), but the same problem occurred. For some reason the `dist` directory is being ignored and not included by npm in the package.

It looks like our machines might have had npm automatically upgraded. Maybe something has changed in npm that is causing it to ignore the `dist` directory now when publishing

Yusang-park:
Thank you. I’ve confirmed that the new version has been released. I’ll continue contributing. If the changes in the PR are unnecessary, feel free to close it!

necolas:
Actually it was caused by this diff, which removed the `files` field from the package.json - https://github.com/facebook/react-strict-dom/commit/959722beff35c297f3bc0295e4d72a27cbd88093 - causing npm to revert to its default behavior of ignoring files in `.gitignore`. What's bizarre is that the `.gitignore` is at the root of the monorepo, but publish is always run in the `react-strict-dom` subdirectory. Might be an npm bug

necolas:
Ok should be fixed in 0.0.32 - https://www.npmjs.com/package/react-strict-dom/v/0.0.32?activeTab=code

Yusang-park:
Interesting. It seems that using files is the most standard approach in the latest version.

PR #269, opened 2025-02-06
Support TypeScript integrations

necolas:
Explicitly add 'types' to each conditional package export.

Close #268


efoken (commented):
This looks good. The types for Web will be used per default and for native one would need a second tsconfig with `"customConditions": ["react-native"]` set in `compilerOptions`

nmn (approved):

necolas:
@efoken ok good to know, I'll add that to the docs. Thanks

PR #294, opened 2025-04-12
allow passing @stylexjs/babel-plugin aliases through react-strict-dom/babel-preset

MoOx:
Until https://github.com/facebook/stylex/pull/856 is merged and released, this is kind of the only option to be able to write imports like defined in a tsconfig file.


necolas:
This just needs to be fixed in stylex

Issue #281, opened 2025-03-11
Running with expo canary


stephenhandley:
### Describe the issue

I've been trying to get react-strict-dom working with expo@canary and running into `Uncaught TypeError: _interopRequireDefault is not a function`.

### Expected behavior

Hoping to be able to use react-strict-dom with Expo latest to port a web ui library which uses React 19 to work with Expo

### Steps to reproduce

Here is a repo with the process I used and should be cloneable to quickly repro:
https://github.com/stephenhandley/expo-rsd

As described there, I followed the process described [here](https://expo.dev/changelog/react-native-78) follow by the process in [your docs](https://facebook.github.io/react-strict-dom/learn/)

### Test case

https://github.com/stephenhandley/expo-rsd

necolas:
This is the stack trace:

```
Uncaught TypeError: _interopRequireDefault is not a function
    at HMRClient.ts:11:1
    at loadModuleImplementation (require.js:277:5)
    at guardedLoadModule (require.js:184:12)
    at metroRequire (require.js:92:7)
    at setupHMR.ts:1:1
    at loadModuleImplementation (require.js:277:5)
    at guardedLoadModule (require.js:184:12)
    at metroRequire (require.js:92:7)
    at effects.ts:8:5
    at loadModuleImplementation (require.js:277:5)
```

Your example doesn't even import RSD in any files that get bundled. I don't see anything pointing to RSD being the issue.

stephenhandley:
The README.md on that repo describes things in more detail, maybe should have put it in this ticket. 

Things work fine going through the canary setup until I follow the install process you're describing in your docs. The postcss and babel configs work fine but the `metro.config.js` causes the above error, this line in particular when uncommented, results in error/stacktrace I shared. 
```
config.resolver.unstable_enablePackageExports = true;
```

I tried workarounds importing RSD components without that but had no luck. Let me know if you would like more details there.

Here's the commit right before doing install process where things render fine with expo canary. 
https://github.com/stephenhandley/expo-rsd/commit/ad3ce6fe6ef912e02a85d0bfb7b46c423eadc5ce

Can you suggest any workarounds for the metro config?

necolas:
> The README.md on that repo describes things in more detail, maybe should have put it in this ticket.

Yeah that would be a good idea. But thanks, that README explains the issue better.

> I then get the following error

```
Web Bundling failed 2384ms index.js (326 modules)
[...]

PR #308, opened 2025-05-17
Add react-strict-dom/postcss-plugin based on @stylexjs/postcss-plugin

javascripter:
# Why
https://github.com/facebook/react-strict-dom/issues/281#issuecomment-2712591645

This PR upstreams [postcss-react-strict-dom](https://github.com/javascripter/postcss-react-strict-dom) to the official repo to make it easier for maintainers to make changes.

I added RSD compatibility to `@stylexjs/postcss-plugin` in [this PR](https://github.com/facebook/stylex/pull/1063), so this adds `react-strict-dom/postcss-plugin` that re-exports the StyleX PostCSS Plugin with the required configuration for RSD consumers.

# How
- Added PostCSS Plugin to `packages/react-strict-dom/postcss/plugin.js`
- Export the plugin as `react-strict-dom/postcss-plugin` in package.json
- Updated `apps/examples` to use the official plugin and removed postcss-react-strict-dom dependency
- Updated docs to use the official plugin
- Updated StyleX to 0.14

# Test Plan
- [x] Run `cd apps/example && npm run dev:web` both from the original repo and this branch and compare side-by-side to make sure output matches 

Left is this branch, right is original
Screenshot 2025-05-20 at 15 12 39

### DIFF
--- a/packages/react-strict-dom/postcss/plugin.js
+++ b/packages/react-strict-dom/postcss/plugin.js
@@ -1,0 +1,41 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+const styleXPlugin = require('@stylexjs/postcss-plugin');
+
+const plugin = ({
+  cwd = process.cwd(),
+  // By default reuses the Babel configuration from the project root.
+  // Use `babelrc: false` to disable this behavior.
+  babelConfig = {},
+  include,
+  exclude
+}) => {
+  include = [
+    // Include the React Strict DOM package's source files by default
+    require.resolve('react-strict-dom'),
+    require.resolve('react-strict-dom/runtime'),
+    ...(include ?? [])
+  ];
+
+  return styleXPlugin({
+    cwd,
+    babelConfig,
+    include,
+    exclude,
+    useCSSLayers: true,
+    importSources: [
+      '@stylexjs/stylex',
+      'stylex',
+      { from: 'react-strict-dom', as: 'css' }
+    ]
+  });
+};
+
+plugin.postcss = true;
+
+module.exports = plugin;
--- a/apps/examples/package.json
+++ b/apps/examples/package.json
@@ -14,7 +14,6 @@
     "expo": "^53.0.11",
     "expo-build-properties": "~0.14.6",
     "expo-status-bar": "~2.2.3",
-    "postcss-react-strict-dom": "^0.0.5",
     "react": "~19.0.0",
     "react-dom": "~19.0.0",
     "react-native": "~0.79.5",
--- a/apps/examples/postcss.config.js
+++ b/apps/examples/postcss.config.js
@@ -6,11 +6,10 @@
  */
 
 module.exports = {
-  plugins: {
-    'postcss-react-strict-dom': {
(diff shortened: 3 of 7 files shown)


packages/react-strict-dom/postcss/plugin.js
```diff
@@ -0,0 +1,41 @@
...
+  cwd = process.cwd(),
+  // By default reuses the Babel configuration from the project root.
+  // Use `babelrc: false` to disable this behavior.
+  babelConfig = {},
+  include,
+  exclude
+}) => {
+  include = [
+    // Include the React Strict DOM package's source files by default
+    require.resolve('react-strict-dom'),
```
necolas, line 22: I think we need to include the runtime export too

packages/react-strict-dom/postcss/plugin.js
```diff
@@ -0,0 +1,41 @@
...
+    ...(include ?? [])
+  ];
+
+  return styleXPlugin({
+    cwd,
+    babelConfig,
+    include,
+    exclude,
+    useCSSLayers: false,
```
necolas, line 31:
```suggestion
    useCSSLayers: true,
```

packages/react-strict-dom/postcss/plugin.js
javascripter: Added in this commit
https://github.com/facebook/react-strict-dom/pull/308/commits/e760f8c94d4bfe39a8f96a237caf4179313c0c1d

packages/react-strict-dom/postcss/plugin.js
javascripter: It seems Expo Web doesn't support CSS Layers. If we enable useCSS Layers and run `npm run dev:web`, it shows the below warning.
I disabled CSS layers for now:
https://github.com/facebook/react-strict-dom/pull/308/commits/b455d438e11581e55d709635e53b27ad5d74740a
```
| Warning: Unknown at rule: @property (src/app/stylex.css:16:10):
|   14 |
|   15 |  @layer priority1{
| > 16 | @property --backgroundColor { syntax: "*"; inherits: false;}
|      |          ^
|   17 | @property --opacity { syntax: "*"; inherits: false;}
|   18 | @property --transform { syntax: "*"; inherits: false;}
|   19 | :root, .xxw3bu5{--primary100-x1ud21w4:black;--primary200-x19lv8z8:white;}
| Warning: Unknown at rule: @property (src/app/stylex.css:17:10):
|   15 |  @layer priority1{
|   16 | @property --backgroundColor { syntax: "*"; inherits: false;}
| > 17 | @property --opacity { syntax: "*"; inherits: false;}
```

necolas:
#311 updates StyleX to 0.13

packages/react-strict-dom/postcss/plugin.js
necolas: Looks like it might be a StyleX bug. Don't think you're allowed to nest `@property` inside `@layer` in CSS

javascripter:
Rebased from main and updated to use `useCSSLayers: true`. Waiting for https://github.com/facebook/stylex/issues/1071 to land so CSS layer warnings will disapper

necolas:
This PR aims to avoid the Expo warnings https://github.com/facebook/stylex/pull/1087

necolas:
We should be able to proceed with this change now that the StyleX upgrade is done

PR #309, opened 2025-05-21
Add name prop to StrictReactDOMInputProps

ksheedlo:
Adds support for input attribute `name`.

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input#name


nitishmehrotra870 (approved):

necolas (changes requested):
Supporting `name` will need more changes than this. I'll do it

PR #310, opened 2025-05-21
Allow "name" attribute for input, select, textarea elements

necolas:
Allows this attribute to be used on web. No support for native at this time.

Close #309


ksheedlo (approved):
LGTM

PR #311, opened 2025-05-21
Update @stylexjs/stylex to 0.13

necolas:
Updates to stylex 0.13 and stubs out the new APIs on native


necolas:
Needs a few more types exported from StyleX, or the old lib path to be put back

mellyeliu (approved):

PR #315, opened 2025-06-11
Upgrade to React 19.0.0 and React Native 0.79.3

necolas:
This diff bumps the dep requirements to React Native 0.79.3 and React 19.0.0 (can't use 19.1.0 with React Native 0.79.3). For now, this diff is unmergable because React Native 0.79.3 has broken Flow type exports.

There is also a change to how React Native is configured to use W3C layout conformance. This can no longer be done automatically by RSD and requires the use of a special prop on the root `html` element of the app, which will render the necessary React Native component behind the scenes. On web, nothing happens.

Some unit tests are also modified to check for RSD-specific warnings and errors, to work around the increase in errors caused by the upgrade to react-test-renderer, which is deprecated for React 19. However, there is no good cross-platform testing alternative at this stage, with the React recommended 3P libraries having incompatible APIs.


MoOx:
🎉 RN 0.79.4 is available and fixes the issue with Flow types

necolas:
0.79.4 is still broken

New pick request for 0.79.5 should fix it https://github.com/reactwg/react-native-releases/issues/1000

PR #324, opened 2025-06-26
Only set userSelect:auto on anchor tags (native)

### LINKED ISSUE #314: [Android] static text components are picked keyboard navigation due to  having `userSelect = "auto"` style by default
### Describe the issue

`h.span` and other text components default to `userSelect: "auto"` style. This causes keyboard navigation (tab or dpad) to pick default text elements as focusable.

https://github.com/user-attachments/assets/df350050-b90b-47bd-ba78-22f572186bd0

React Native's `Text` behaves correctly (is skipped by keyboard nav).

How this happens:
- this starts with default style of `userSelect: "auto"` [here](https://github.com/facebook/react-strict-dom/blob/a1f618fa00511021f61e902759f38add07d9675a/packages/react-strict-dom/src/native/modules/createStrictDOMTextComponent.js#L48)
- which causes `selectable: true` prop passed to React Native component [here](https://github.com/facebook/react-native/blob/198c6fc81cda67a46fff457cc39eb2a297eddc01/packages/react-native/Libraries/Text/Text.js#L152)
- then this sets `textIsSelectable` attribute on `ReactTextView` which is descendant of Android `TextView` [here](https://github.com/facebook/react-native/blob/198c6fc81cda67a46fff457cc39eb2a297eddc01/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/views/text/ReactTextView.java#L539)
- finally Android's TextView sets couple of flags: `focusable`, `focusableInTouchMode`, `clickable`, and `longClickable` to true [docs](https://developer.android.com/reference/android/widget/TextView#setTextIsSelectable(boolean))

Since React Native is not exposing any of the `focusable`, etc flags, this cannot be overridden, to make text non-focusable.

NOTE:
- this is different from screen reader behavior when it's expected that text elements are picked up, this issue is for keyboard navigation.
- on iOS these default text components behave correctly
- on iOS these default text components aren't do not have "selectable" text but rather enable Copy context action

### Expected behavior

Default text components should not be picked by keyboard/dpad navigation. 

IMO reasonable solution would be to have default to `userSelect: none` for native environments (iOS, Android), as this matches the default behavior on both OSes.

### Steps to reproduce

1. In Android emulator run the test case repo
2. Use Tab key (or dpad) to cycle through elements

### Test case

https://github.com/mdjastrzebski/repro-rsd-keyboard-navigation

### Additional comments

_No response_

necolas:
Fix #314

### DIFF
--- a/packages/react-strict-dom/src/native/modules/createStrictDOMTextComponent.js
+++ b/packages/react-strict-dom/src/native/modules/createStrictDOMTextComponent.js
@@ -18,7 +18,6 @@ import { errorMsg } from '../../shared/logUtils';
 import { mergeRefs } from '../../shared/mergeRefs';
 import { useNativeProps } from './useNativeProps';
 import { useStrictDOMElement } from './useStrictDOMElement';
-import * as stylex from '../stylex';
 
 type StrictProps = $ReadOnly<{
   ...StrictPropsOriginal,
@@ -31,7 +30,7 @@ function hasElementChildren(children: mixed): boolean {
 
 export function createStrictDOMTextComponent<T, P: StrictProps>(
   tagName: string,
-  _defaultProps?: P
+  defaultProps?: P
 ): component(ref?: React.RefSetter<T>, ...P) {
   const component: React.AbstractComponent<P, T> = React.forwardRef(
     function (props, forwardedRef) {
@@ -43,10 +42,6 @@ export function createStrictDOMTextComponent<T, P: StrictProps>(
       /**
        * Resolve global HTML and style props
        */
-
-      const defaultProps = {
-        style: [_defaultProps?.style, styles.userSelectAuto]
-      };
 
       const { customProperties, nativeProps, inheritableStyle } =
         useNativeProps(defaultProps, props, {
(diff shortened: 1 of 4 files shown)


mdjastrzebski (commented):
I've done some practical checks, and:
- on iOS: neither `onPress` nor `userSelect` style triggers keyboard navigation focus. Wrapping `Text` in `Pressable` (or simply `<View accessible>`) allows for focusing the text. However, when nesting that inside `Text` you get whacky layout.

CleanShot 2025-06-27 at 15 08 01@2x

- on Android: having `onPress` prop alone is sufficient for `Text` component to receive keyboard navigation focus (being wrapped in `Pressable` is also fine)

Therefore, I would recommend on not applying `userSelect: auto` to `a` elements, as:
- it's not needed on Android, having `onPress` prop is enough
- it does not work on iOS anyway
- AFAIU the purpose of `userSelect` style is to allow for user to copy given text, not to make it focusable by keyboard, that would be `tabIndex`/`focusable` props (but these work only on Android anyway)

You can check the experiements repo here: https://github.com/mdjastrzebski/repro-rsd-keyboard-navigation

PR #337, opened 2025-07-11
Add native support for `css.types.*` functions

### LINKED ISSUE #331: Missing `css.types` Polyfill in Native `stylex` Export Blocks Cross-Platform Usage
### Description

We're using `react-strict-dom` (v0.0.39) to enable theming for web and react-native apps. After upgrading, we began using `css.types` with `defineVars` to help enforce type safety and linting (per [stylex docs](https://stylexjs.com/docs/api/javascript/types/)). The eslint plugin for stylex encourages this usage.

### Problem

On native, `stylex` is polyfilled, but the exported `css` object does not include a polyfill for `css.types.*`.

- `css.types` is undefined on native and breaks code that expects it.
- The eslint plugin throws errors when using string/int values without `css.types`, but there’s no way to fix this due to the missing polyfill.

**Relevant lines and files:**

- [`index.js` export](https://github.com/facebook/react-strict-dom/blob/main/packages/react-strict-dom/src/native/index.js#L55):

 ```js
  const css: TStyleX = stylex as $FlowFixMe;
```

In the `dist/native/index.js` this resolves to:

```js
var stylex = /*#__PURE__*/Object.freeze({
  __proto__: null,
  __customProperties: __customProperties,
  create: create,
  createTheme: createTheme,
  defineConsts: defineConsts,
  defineVars: defineVars,
  firstThatWorks: firstThatWorks,
  keyframes: keyframes,
  positionTry: positionTry,
  props: props
});
```

Here, stylex/css does not have a types property on native.

### Request for Guidance

What is the recommended path forward for unified cross-platform usage?

- Should we polyfill css.types on native (e.g., returning the string/int values as on web), or is there a better approach?
- Are there best practices for handling theming and typing for native with stylex and react-strict-dom?

### Additional Notes:

- The eslint plugin is very helpful for enforcing best practices, but the current gap with css.types on native is a blocker.
- Our goal is a unified codebase for web and native.

necolas:
These functions pass the value straight through on native.

Fix #331

### DIFF
--- a/apps/examples/src/components/tokens.stylex.js
+++ b/apps/examples/src/components/tokens.stylex.js
@@ -19,7 +19,7 @@ export const tokens: StyleVars<
     inputPlaceholderColor: string
   }>
 > = css.defineVars({
-  squareColor: 'red',
+  squareColor: css.types.color('red'),
   textColor: {
     default: 'darkred',
     '@media (prefers-color-scheme: dark)': 'lightred'
--- a/packages/react-strict-dom/src/native/stylex/index.js
+++ b/packages/react-strict-dom/src/native/stylex/index.js
@@ -641,3 +641,78 @@ export function props(
 
   return nativeProps;
 }
+
+type ValueWithDefault<+T> =
+  | T
+  | $ReadOnly<{
+      default: ValueWithDefault<T>,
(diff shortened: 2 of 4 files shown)


christianbach:
LGTM!

PR #343, opened 2025-07-23
(docs) add Next.js setup instructions

MoOx:
There are some diff between Expo & Next.js, to setup this properly (especially because of babel config - when using turbopack), that's why I think it's important for new comers to have this in the guide.

### DIFF
--- a/apps/website/docs/learn/02-environment-setup.md
+++ b/apps/website/docs/learn/02-environment-setup.md
@@ -6,6 +6,9 @@ slug: /learn/setup
 
 <p className="text-xl">Learn how to configure the environment needed to use React Strict DOM.</p>
 
+* [Expo framework](#expo-framework) - Configure React Strict DOM in an Expo app.
+* [Next.js framework](#nextjs-framework) - Configure React Strict DOM in a Next.js website using App Router.
+
 ## Expo framework
 
 [Expo](https://expo.dev/) is a production-grade, cross-platform React framework that is the recommended solution for creating apps with React Strict DOM. The instructions in the rest of this guide are tailored to Expo, but can be adapted by readers to work with other frameworks.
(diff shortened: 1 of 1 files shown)


apps/website/docs/learn/02-environment-setup.md
```diff
@@ -147,8 +150,174 @@ export function App() {
...
+      "*.{js,jsx,ts,tsx}": {
+        loaders: [getBabelLoader()],
+      },
+    },
+  },
+
+  webpack: (config, { webpack }) => {
+    config.resolve.mainFields = ["module", "main"];
+
+    config.module.rules.push({
```
javascripter, line 253: We should add `exclude` option in the module rule so that unrelated node_modules should not go through this babel loader. 
```typescript
        exclude: /node_modules(?!\/react-strict-dom)/,
```

See:
https://github.com/facebook/stylex/blob/4a4fe98486cf99a9783e10ec07a41452790a17a6/examples/example-nextjs/next.config.js#L21

apps/website/docs/learn/02-environment-setup.md
```diff
@@ -147,8 +150,174 @@ export function App() {
...
+When using Next.js App Router, babel is not the default compiler, but can still be used.
+Create or modify a `babelLoader.config.js` file as follows (note that it's not a `babel.config.js`). This is used to optimize builds and enables static extraction of CSS for web. Learn how to configure the [babel-preset](/api/babel-preset/) in the API docs.
+
+```js title="babelLoader.config.js"
+import path from "path";
+
+const config = {
+  presets: [
+    [
+      "next/dist/compiled/babel/preset-typescript",
```
javascripter, line 170:
```
      "next/dist/compiled/babel/preset-typescript",
```

This line might not be necessary. We (not Meta) started using Turbopack for development at our company and so far we are using the below config without that line.

```typescript
function getReactStrictDOMBabelLoader() {
  const dev = process.env.NODE_ENV !== 'production'

  return {
    loader: 'babel-loader',
    options: {
      cacheDirectory: true,
      parserOpts: {
        plugins: ['typescript', 'jsx'],
      },
      presets: [
        [
          'react-strict-dom/babel-preset',
          {
            debug: dev,
            dev,
            rootDir: process.cwd(),
          },
        ],
      ],
    },
  }
}

function withReactStrictDOM(nextConfig: NextConfig) {
  return {
    ...nextConfig,
    webpack(config: any, context: WebpackConfigContext) {
      // Configure StyleX / React Strict DOM
      config.module.rules.push({
        test: /\.(js|jsx|ts|tsx)$/,
        exclude: /node_modules(?!\/react-strict-dom)/,
        use: [getReactStrictDOMBabelLoader()],
      })

      if (typeof nextConfig.webpack === 'function') {
        return nextConfig.webpack(config, context)
      }

      return config
    },
    turbopack: {
      ...nextConfig.turbopack,
      rules: {
        ...nextConfig.turbopack?.rules,
        // Note: Intentionally excluding top-level files in the src directory to skip middleware.ts from being processed
        // due to a Turbopack bug (https://github.com/vercel/next.js/issues/79592)
        './src/*/**/*.{js,jsx,ts,tsx}': {
          default: {
            foreign: false,
            loaders: [getReactStrictDOMBabelLoader()],
          },
        },
        '**/node_modules/react-strict-dom/**/*.{js,jsx,ts,tsx}': {
          default: {
            foreign: false,
            loaders: [getReactStrictDOMBabelLoader()],
          },
        },
      },
    },
  }
}
```

Note that there is currently a bug in Next.js that prevents Turbpack loader from running when `middleware.ts` exists.
We are working around this limitation by excluding top-level `src/*.{ts,tsx}` files from being compiled with babel-loader, but this is specific to our and probably isn't something we want to recommend here.

```typescript
foreign: true
```
is meant to exclude node_modules etc but I'm not too sure the exact behavior as it's not documented well. Just something we added and it seems to work.

necolas:
Merged in #368. We can iterate from here as needed

PR #361, opened 2025-08-04
Go back to using ref callback instead of useImperativeHandle

necolas:
We shouldn't create an object if the underlying node is null, as this handle is meant to emulate the node but can't proxy methods if none exist.

### DIFF
--- a/packages/react-strict-dom/src/native/modules/useStrictDOMElement.js
+++ b/packages/react-strict-dom/src/native/modules/useStrictDOMElement.js
@@ -7,9 +7,12 @@
  * @flow strict-local
  */
 
+import type { CallbackRef } from '../../types/react';
+
 import * as React from 'react';
 
 import { errorMsg } from '../../shared/logUtils';
+import { useElementCallback } from '../../shared/useElementCallback';
 import { useViewportScale } from './ContextViewportScale';
 
 type Options = {
@@ -25,9 +28,8 @@ function errorUnimplemented(name: string) {
   }
 }
 
-function proxy(nativeRef: React.RefObject<Node>, name: string) {
+function proxy(node: Node, name: string) {
   return (...args: Array<mixed>) => {
-    const node = nativeRef.current;
     if (node?.[name]) {
       return node[name](...args);
     }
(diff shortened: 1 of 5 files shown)


packages/react-strict-dom/src/native/modules/useStrictDOMElement.js
```diff
@@ -43,62 +43,65 @@ export function useStrictDOMElement<T>(
...
-      scroll: proxy(nativeRef, 'scroll'),
-      scrollBy: proxy(nativeRef, 'scrollBy'),
-      scrollIntoView: proxy(nativeRef, 'scrollIntoView'),
-      scrollTo: proxy(nativeRef, 'scrollTo'),
-      select: proxy(nativeRef, 'select'),
-      setSelectionRange: proxy(nativeRef, 'setSelectionRange'),
-      setPointerCapture: proxy(nativeRef, 'setPointerCapture'),
-      showPicker: proxy(nativeRef, 'showPicker')
-    };
+    return nativeRef.current == null
```
yungsters, line 46: This won't be re-evaluated when`nativeRef.current` changes.

Currently, this imperative handle will only be re-evaluated when `tagName` or `viewportScale` change, because they're enumerated as dependencies.

You *could* store `nativeRef.current` in state, but then you'll be updating this component only to invalidate `useImperativeHandle`, which is unnecessarily inefficient.

A better solution would be to use a callback ref instead of `useImperativeHandle`, which allows you to update `ref` without creating any new state.

packages/react-strict-dom/src/native/modules/useStrictDOMElement.js
```diff
@@ -38,68 +40,89 @@ function proxy(nativeRef: React.RefObject<Node>, name: string) {
...
-      blur: proxy(nativeRef, 'blur'),
-      click: proxy(nativeRef, 'click'),
-      get complete() {
-        if (tagName === 'img') {
-          const node = nativeRef.current as Node;
-          if (node?.complete == null) {
-            // Assume images are never pre-loaded in React Native
-            return false;
+  const elementCallback = useElementCallback(
+    React.useCallback(
```
yungsters, line 47: I don't think you need `useElementCallback` here.

The `useCallback` already has `ref` as a dependency, which means the function returned by `useCallback` will already be invalidated if `ref` changes. And when that new function is passed into the native component's `ref` prop below, it will naturally detach the previous one and attach the new one.

packages/react-strict-dom/src/native/modules/useStrictDOMElement.js
```diff
@@ -38,68 +40,89 @@ function proxy(nativeRef: React.RefObject<Node>, name: string) {
...
+      // $FlowFixMe[unclear-type]
+      (node: any) => {
+        const obj = {
```
yungsters, line 50: You should memoize this `obj`.

There are multiple scenarios in which the same instance is passed to multiple refs:

- An enclosing `Activity` becomes hidden, then visible. (DOM nodes do not get deallocated when `Activity` becomes hidden, so when `Activity` becomes visible, that same DOM node would be reattached to refs.)
- A new ref is supplied to a native component that is already mounted. (The old ref and the new ref would receive instances that are `===`.)

PR #368, opened 2025-08-11
Add Next.js setup instructions to docs

necolas:
Modified version of #343

* Creates separate Expo/Next docs pages.
* Incorporates feedback on #343 

cc @MoOx @javascripter

### DIFF
--- a/apps/website/docs/learn/environment-setup/01-expo.md
+++ b/apps/website/docs/learn/environment-setup/01-expo.md
@@ -2,11 +2,11 @@
 slug: /learn/setup
 ---
 
-# Environment setup
+# Expo
 
-<p className="text-xl">Learn how to configure the environment needed to use React Strict DOM.</p>
+<p className="text-xl">Learn how to configure Expo to use React Strict DOM.</p>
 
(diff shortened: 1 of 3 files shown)


MoOx (commented):
Thanks for handling this !

I quickly checked and see that you used `babel.config.js` file name. I should have mentioned it, but using this name will make next to tell you that you cannot use turbopack because "babel.config.js" isn't compatible with it.

```
 ⨯ You are using configuration and/or tools that are not yet
supported by Next.js with Turbopack:
Babel detected (babel.config.js)
  Babel is not yet supported. To use Turbopack at the moment,
  you'll need to remove your usage of Babel.

If you cannot make the changes above, but still want to try out
Next.js with Turbopack, create the Next.js playground app
by running the following commands:

  npx create-next-app --example with-turbopack with-turbopack-app
  cd with-turbopack-app
  npm run dev

 ⚠ Learn more about Next.js and Turbopack: https://nextjs.org/docs/architecture/turbopack
 ```

So that's why I used a different name. It should be explained otherwise following this guide as it is will throw the error above.

PR #376, opened 2025-08-20
Change the accepted values of `<compat.native>` prop `as`

necolas:
This prop previously mixed the semantics of web and native in the accepted values. Now the values match the tag names on web, and indicate the equivalent `html.*` functional role that the compat.native component performs. LLMs have a better success rate using this API after this change. So it was likely somewhat confusing for people too.


necolas:
Reviewed/approved by Ken over chat

PR #378, opened 2025-08-24
Exclude default styles from `<compat.native>`

necolas:
* Don't include web defaults of `position:static` and `boxSizing:content-box` in the transformed styles when using `<compat.native>`. Simply transform the input styles.
* Don't bother including web defaults on `<br />`.


ksheedlo (approved):
LGTM

PR #379, opened 2025-08-26
Configure hermes parser not to output fowardRef wrapper

necolas:
Target React 19 runtime so we avoid the forwardRef wrapper and props object copying overhead.

What the component syntax compiles to now

image


KingAmo:
we are still in react 18, should we stay on old version of RSD ?

PR #383, opened 2025-09-02
Fix Next.js installation instructions

### LINKED ISSUE #382: Next.js installation guide not working (repo included)
### Describe the issue

Hi 👋

I followed the Next.js installation guide for react-strict-dom, but the setup doesn’t work and I get errors.

To make it easier to reproduce, I created a minimal repo that follows the docs exactly:
👉 https://github.com/adamkrim/nextjs-react-strict-dom

Could you confirm if the instructions are up to date, or share a working Next.js example?

Thanks!

### Expected behavior

Following the instructions, the app should run with styles applied.

### Steps to reproduce

Follow the instructions of the doc from a fresh nextjs install, or clone the repo: https://github.com/adamkrim/nextjs-react-strict-dom
1. `pnpm i`
2. `pnpm run dev`

### Test case

https://github.com/adamkrim/nextjs-react-strict-dom

### Additional comments

_No response_

MoOx:
Closes #382 

That was my initial config in babel but think we adjusted at some point.

I just started a new next.js app and faced the same issues as mentioned in #382. With the changes in this PR it's working without trouble for webpack and turbopack.

Cc @javascripter 

Ref https://github.com/facebook/react-strict-dom/pull/343#discussion_r2245468263

### DIFF
--- a/apps/website/docs/learn/environment-setup/02-next.md
+++ b/apps/website/docs/learn/environment-setup/02-next.md
@@ -66,17 +66,14 @@ const config = {
       babelConfig: babelLoader,
       useLayers: true,
     },
-    autoprefixer: {},
+    // If your pipeline doesn't already do CSS auto-prefixing, feel free to install as well autoprefixer
+    // $ npm install --dev autoprefixer
+    // and uncomment the line below
+    // autoprefixer: {},
   },
 };
 
 export default config;
-```
-
-If necessary, run
-
-```
-npm install --dev autoprefixer
 ```
 
 ## Next.js configuration
--- a/apps/website/docs/learn/environment-setup/02-next.md
+++ b/apps/website/docs/learn/environment-setup/02-next.md
@@ -20,11 +20,13 @@ Babel is not the default compiler when using Next.js App Router, but it can stil
 const dev = process.env.NODE_ENV !== 'production'
 
 const config = {
-  cacheDirectory: true,
-  parserOpts: {
-    plugins: ['typescript', 'jsx'],
-  },
   presets: [
+    [
+      "next/dist/compiled/babel/preset-typescript",
+      {
+        allowNamespaces: true,
+      },
+    ],
     [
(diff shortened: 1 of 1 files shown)


necolas:
Could autoprefixer be installed by the RSD postcss plugin to avoid users having to install it separately when using RSD?

javascripter:
@necolas 

> Could autoprefixer be installed by the RSD postcss plugin to avoid users having to install it separately when using RSD?

1. I don't think there's an easy way to compose and call other PostCSS Plugins in RSD PostCSS Plugin itself
    - https://postcss.org/docs/writing-a-postcss-plugin
2. RN/Expo community is moving towards supporting isolated dependencies (pnpm etc), so direct dependencies listed in postcss.config.js should be installed by the user explicitly going forward
    - https://kitten.sh/blog/autolinkings-broken-promise

So, adding it to the docs makes more sense I think. It's also common to instruct users to install autoprefixer [in the docs](https://docs.expo.dev/guides/tailwind/#configuration).

javascripter:
> My perspective is that requiring people to install multiple packages to make things work is not a great user experience. That's why I'm wondering if there's a way to avoid needing autoprefixer be installed by users of RSD.

I realized we could just pass CSS string generated by stylex babel plugin through autoprefixer/lightningcss JS API [directly](https://github.com/facebook/stylex/blob/648008d81d58d5caa2beffd9699ce2e23f4d2480/packages/%40stylexjs/postcss-plugin/src/plugin.js#L78) in stylex postcss plugin. That avoids requiring users to install autoprefixer themselves and users get automatic vendor-prefixing.

Tailwind v4 uses Lightning CSS under the hood to [do automatic vendor-prefixing](https://tailwindcss.com/blog/tailwindcss-v4#simplified-installation) without requiring other dependencies, so it'll be similar to that.

That said, as autoprefixer isn't a hard requirement, it also makes sense to comment it out and makes it optional. 
What do you think?

PR #384, opened 2025-09-06
Add native polyfill for willChange

vincentriemer:
Add native polyfill for willChange

At least on RN Android there's a specific prop, `renderToHardwareTextureAndroid`, that promotes an RN view to a GPU layer which is almost exactly the same responsibility of the web's `will-change` CSS property so this diff maps the two properties together.


necolas:
Rebased and merged 9a852008fcb24f9e1e57cabe96f2dbf4a91dd089

PR #389, opened 2025-09-16
Minor optimization to display block shim

necolas:
Avoid setting styles if they are already the default for View. The cost of each style property in React Native is relatively high, so this micro optimization can still add up.


packages/react-strict-dom/src/native/modules/createStrictDOMComponent.js
```diff
@@ -135,13 +135,25 @@ export function createStrictDOMComponent<T, P: StrictProps>(
...
+        nativeProps.style.alignItems = 'stretch';
+      }
+      if (nativeProps.style.flexBasis != null) {
+        nativeProps.style.flexBasis = 'auto';
+      }
+      if (nativeProps.style.flexDirection != null) {
+        nativeProps.style.flexDirection = 'column';
+      }
+      if (nativeProps.style.flexShrink != null) {
+        nativeProps.style.flexShrink = 0;
```
necolas, line 149: This might benefit from a little more logic, because I think flexShrink has an effect if the element is block but also a flex child

PR #391, opened 2025-09-16
Apps/examples refactor

necolas:


necolas:
I love how much Expo takes care of by default, even for the web setup.

necolas:
#396 updates the apps to Expo SDK 54

PR #393, opened 2025-09-18
Move postcss plugin back into monorepo

necolas:
Allows for configuration to be simplified, and avoids confusing need to mix react-strict-dom and stylex names.

### DIFF
--- a/packages/postcss-react-strict-dom/src/bundler.js
+++ b/packages/postcss-react-strict-dom/src/bundler.js
@@ -1,0 +1,74 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+const babel = require('@babel/core');
+const stylexBabelPlugin = require('@stylexjs/babel-plugin');
+
+// Creates a stateful bundler for processing styles using Babel.
+module.exports = function createBundler() {
+  const styleXRulesMap = new Map();
+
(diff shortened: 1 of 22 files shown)


necolas:
Just @necolas for now please. Thanks!

packages/postcss-react-strict-dom/src/bundler.js
```diff
@@ -0,0 +1,74 @@
...
+const babel = require('@babel/core');
+const stylexBabelPlugin = require('@stylexjs/babel-plugin');
+
+// Creates a stateful bundler for processing styles using Babel.
+module.exports = function createBundler() {
+  const styleXRulesMap = new Map();
+
+  // Determines if the source code should be transformed
+  function shouldTransform(sourceCode) {
+    return sourceCode.includes('react-strict-dom');
```
javascripter, line 17: Is there an edge case where some files without react-strict-dom string must still be transformed due to variables?

Something like the below:

src/index.ts:
```typescript
import { colors } from './vars.stylex';
import { Text } from './text';

// No direct react-strict-dom imports, but still needs transformations due to vars?

export default function App() {
  return <Text color={colors.primary}>Hello</Component>
}
```

src/text.tsx:
```typescript
import { html , css } from 'react-strict-dom';
export function Text({ color, style, ...props} ) { 
  return <html.div style={[styles.text(color), style]} {...props} />
}

const styles = css.create({
  text: (color) => ({
    color
  })
})
```

src/vars.stylex.ts:
```typescript
import { css } from 'react-strict-dom'

export const colors = stylex.defineVars({
  primary: 'green'
});
```

PR #399, opened 2025-09-30
add react-strict-animated library

vincentriemer:
add react-strict-animated library

This introduces a new package to the RSD repo that provides a subset of React Native's API that works on both web and native, and built with integration of react-strict-dom in mind. The native implementation is largely a thin passthrough to the proper Animated API and some settings pre-configured (such as useAnimatedDriver always being enabled).

The web side is a new Animated implementation that ends up being driven by the Web Animations API.

### DIFF
--- a/packages/react-strict-animated/src/native/components.js
+++ b/packages/react-strict-animated/src/native/components.js
@@ -1,0 +1,166 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow strict-local
+ */
+
+/* eslint-disable no-unreachable */
+
+import type { AnimatedStyleValue } from '../shared/SharedAnimatedTypes';
+import type { Text, View } from 'react-native';
+import type { ImageProps } from 'react-native/Libraries/Image/ImageProps';
+import type { html } from 'react-strict-dom';
+
+import * as React from 'react';
+import { Animated } from 'react-native';
+
+import { compat, css, useViewportScale_DO_NOT_USE } from 'react-strict-dom';
+
+const styles = css.create({
+  defaults: {
+    boxSizing: 'content-box',
+    position: 'static'
+  }
+});
+
+function useViewportScaledAnimatedStyle(
+  animatedStyle: ?AnimatedStyleValue<Animated.Node>
+): ?AnimatedStyleValue<Animated.Node> {
+  const { scale: viewportScale } = useViewportScale_DO_NOT_USE();
+
+  return React.useMemo(() => {
+    if (animatedStyle == null || viewportScale === 1) {
+      return animatedStyle;
+    }
+
+    const transforms = animatedStyle.transform;
+    if (transforms == null) {
+      return animatedStyle;
+    }
+
+    let didScaleTransforms = false;
+    const scaledTransforms = transforms.map((transform) => {
+      if (transform?.translateX != null) {
+        const translateX = transform.translateX;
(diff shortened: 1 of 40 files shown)


necolas:
Thanks for bringing the animation library you've made to the OSS project! Want a hand with the build and package parts?

vincentriemer:
I've tried digging into how to resolve the type generation issues but haven't had any luck yet :/

packages/react-strict-animated/src/web/nodes/AnimatedInterpolation.js
```diff
@@ -0,0 +1,298 @@
...
+
+const numericComponentRegex = /[+-]?(?:\d+\.?\d*|\.\d+)(?:[eE][+-]?\d+)?/g;
+
+// Maps string inputs an RGBA color or an array of numeric components
+function mapStringToNumericComponents(
+  input: string
+):
+  | { components: [number, number, number, number], isColor: true }
+  | { components: $ReadOnlyArray<number | string>, isColor: false } {
+  let normalizedColor = normalizeColor(input);
```
necolas: Is there any way we could avoid the need for `normalizeColor`?

packages/react-strict-animated/src/native/components.js
```diff
@@ -0,0 +1,108 @@
...
+      {...htmlProps}
+      as="div"
+      ref={ref}
+      style={[styles.defaults, htmlProps.style]}
+    >
+      {(nativeProps: React.PropsOf<View>) => {
+        return (
+          <Animated.View
+            {...nativeProps}
+            style={[nativeProps.style, animatedStyle]}
```
necolas: Can `animatedStyle` be passed to `compat.native` so it picks up any relevant polyfills?

necolas:
Hey @vincentriemer, what's the plan here? Did you make any progress on incorporating the viewport scaling polyfill into the animated values?

vincentriemer:
I mean the other option to support it would be to have the package track viewport scaling all on its own, and yeah the internal version isn't taking viewport scaling into account yet I've been focusing on this version first. 

`normalizeColor` is something we technically aren't using as currently I haven't enabled any color-based properties to be animatable but if we do ever add any color-based properties `normalizeColor` is important because interpolation is fundamentally done on numeric values, but that's not really how we represent colors in RN/web, so if you want to consistently interpolate between different colors its easier to first normalize them to consistent representation, in this case an rgba array of numbers. 

FWIW if I could I would just have this implementation point to the `normalizeColor` that lives in RN but that's an internal file I can't reference so the one that's in this codebase is more or less a duplicate of it.

necolas:
> if I could I would just have this implementation point to the normalizeColor that lives in RN 

We have a separate package for it these days I believe: https://www.npmjs.com/package/@react-native/normalize-colors

---

For the current approach we could export useViewportScale prefixed with `unstable_` or `DO_NOT_USE_` to avoid it looking like "blessed" public API. Don't want to hold this up on figuring out something "ideal".

I don't have any great ideas for a clean relationship between packages. This is my understanding of the options and tradeoffs. Sharing in case anyone has suggestions. 

**Part of RSD package**
* Difficulty importing from `react-strict-dom` rather than relative imports
* Exported as subpath: `react-strict-dom/animated`
* Can share RN imports and other internals with RSD package
* No need to export internals / "temporary" workarounds like useViewportScale

**New package**
* Built only on public API of RSD
* Exported as separate new package: `react-strict-animated` with RSD dependency
* Can share imports or internals with RSD
* Needs to modify RSD exports to include internals
* Can't potentially be incorporated into RSD babel transform

MoOx:
Do you have an idea when this will be available to try it out ?

PR #410, opened 2025-10-02
Export StrictClickEvent type from StrictReactDOMProps

MoOx:
This is useful when defining a callback, at least for typescript (otherwise, you need to redefined explicitly the type that match what a onClick could expect).

### DIFF
--- a/packages/react-strict-dom/src/types/StrictReactDOMProps.js
+++ b/packages/react-strict-dom/src/types/StrictReactDOMProps.js
@@ -173,7 +173,7 @@ type SyntheticEvent<+T> = $ReadOnly<{|
 |}>;
 */
 
-type StrictClickEvent = $ReadOnly<{|
+export type StrictClickEvent = $ReadOnly<{|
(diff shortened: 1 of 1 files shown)


javascripter:
FWIW for TypeScript you can do something like the below:
```typescript
type StrictClickEventHandler = NonNullable<React.ComponentProps<typeof html.div>['onClick']>

// if you want to get the click event type itself
type StrictClickEvent =  Parameters<StrictClickEventHandler>[0]
// or 
type StrictClickEvent =  StrictClickEventHandler extends (event: infer E) => any ? E : never
```

`React.ComponentProps` works well as long as the prop types do not rely on generics (e.g. `type Props<T> = { onFoo: (arg: T) => void }`), meaning it works fine for RSD.

necolas:
I think this type is a leftover from the early days when we had types like StrictHTMLElement. So it might be best to first FlowFixMe this event. Later we can properly type the event props to match what's expected on web

PR #411, opened 2025-10-08
Add llms.txt to docs

necolas:
Add an initial LLM context to help bootstrap LLM-based development. We've had success with this approach internally

### DIFF
--- a/packages/website/docs/api/02-css/01-create.md
+++ b/packages/website/docs/api/02-css/01-create.md
@@ -4,7 +4,7 @@
 
 :::warning
 
-On web the style compiler and static types do not yet enforce constraints on the properties and values that can be used (e.g., various short-form properties are disallowed). This is a work in progress that depends on improvements to StyleX.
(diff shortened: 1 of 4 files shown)


joaqo:
@necolas Would you consider having this file also be sent alongside the rsd npm node module? I added the link to llms.txt you show in your docs to my claude code skills, and it works great, but it occurred to me this approach has 2 disadvantages:

- Possible version missmatch, the llms doc in https://facebook.github.io/react-strict-dom/llms.txt I assume will point to the latest version of rsd, but the one I have locally installed, and that claude should code for, may be older. Having it be tied to my installed node package would ensure they always match.
- It would be a bit faster, being a local file and all.

The implementation user side would be simple, I could just tell my claude code skill to learn how to use rsd by looking at `node_modules/react-strict-dom/blah/blah/llms.txt`.

necolas:
Yes for sure, I think we should add a skill to the npm package. I've been advocating for React to ship skills in the npm package too - help with migrations and use modern patterns instead of whatever is the default tendency of LLMs

joaqo:
Thats great! I'm absolutely in love with rsd btw, moving our stack to it atm, thanks for finally solving rn + web 👏

PR #416, opened 2025-10-21
Feature/delegate to first font in fallback list on React Native

Jta26:
In working with some internal stuff, I needed to do a pretty gnarly code-split between Web, Android, and iOS due to different platforms having different names for their fonts. In addition, I needed to do `!important` on web to override a bunch of high specificity styles in a global CSS file. When I added !important, I noticed the incompatibility here. 

`!important` is being sent down through an `html.span` to the underlying `<Text>`'s TextStyleProps in RN, resulting in an undefined font family. RSD will also send down fallback fonts, also resulting in undefined fonts. Ideally `!important` is never used in atomic styles, but in the event that it is used, I think it's reasonable to strip it off completely.

Example:
image

Thought I'd try to throw up a PR that fixes it.

Two Questions
1. Should we change this to *always* strip off !important for all CSS properties? There is no value in doing anything with `!important` on native imo.
2. Is delegating to the first font in the list of fallbacks the *right* thing to do here? Should we do a runtime check on font's somehow to support proper fallbacks?

Snapshot tests written by AI

### DIFF
--- a/packages/react-strict-dom/tests/__snapshots__/css-test.native.js.snap
+++ b/packages/react-strict-dom/tests/__snapshots__/css-test.native.js.snap
@@ -1,5 +1,47 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`properties: custom property css.createTheme: css.__customProperties 1`] = `
+{
+  "rootColor__id__1": "red",
+  "rootColor__id__3": "red",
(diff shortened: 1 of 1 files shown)


Jta26:
Synced offline, updated this PR to not do anything with `!important`, but it will still handle delegating to the first font in the list of fallbacks on React Native.

PR #436, opened 2025-12-10
Handle  `autoCapitalize` for native `<TextInput />`

wlabu:
The `autoCapitalize` prop is not being forwarded through to the native input component. Which means that the following has no effect on the mobile device software-keyboard when the input is focused:

```tsx
<html.input autoCapitalize="none" />
```

This change maps the `autoCapitalize` prop's various values to the underlying native `<TextInput />` element, and allows the software-keyboard to react to the prop.

---

There is some complexity with mapping the prop's values since the types declare a list larger than what is supported in react-native.

The declared type's values are:

```ts
"none" | "off" | "on" | "sentences" | "words" | "characters"
```

The list of supported values in react-native is:

```ts
"none" | "sentences" | "words" | "characters"
```

Further, there was also an existing test which assumed that the `autoCapitalize` prop could be set to boolean `true`, as follows:

```tsx
<html.input autoCapitalize />
```

The way this PR maps these values has been based on the MDN docs:
- https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/autocapitalize

|prop value|native value|
|---|---|
|none|none|
|off|none|
|`true`|sentences|
|on|sentences|
|sentences|sentences|
|words|words|
|characters|characters|

---

<details>
<summary>👋 Quick disclaimer</summary>

> RSD is very exciting, and has been a lot of fun to use.
> Thank you very much for the hard work!
> 
> This is my first PR here, and even though I _think_ I've covered most of the requirements and tests, I'd be happy to amend it however required.

</details>

### DIFF
--- a/packages/react-strict-dom/src/types/renderer.native.js
+++ b/packages/react-strict-dom/src/types/renderer.native.js
@@ -55,8 +55,8 @@ type ReactNativeProps = {
     text?: ?string
   },
   accessibilityViewIsModal?: ?boolean,
-  autoComplete?: ?string,
   autoCapitalize?: ?('none' | 'sentences' | 'words' | 'characters'),
+  autoComplete?: ?string,
   alt?: ?Stringish,
   animated?: ?boolean, // non-standard
   caretHidden?: TextInputProps['caretHidden'],


packages/react-strict-dom/src/types/StrictReactDOMProps.js
```diff
@@ -281,6 +281,7 @@ export type StrictReactDOMProps = $ReadOnly<{
 
   // Other
   autoCapitalize?: ?(
+    | true
     | 'off'
     | 'none'
     | 'on'
```
necolas: Potentially we could remove `off` and `on` as they were part of the pre-standard, iOS version of autoCapitalize and were retained for backwards compatibility. Not necessary for now, just putting it out there.

packages/react-strict-dom/tests/html/html-test.native.js
```diff
@@ -392,6 +392,25 @@ describe('<html.*> (native polyfills)', () => {
...
+      test('"autoCapitalize" prop', () => {
+        [
+          // web only
+          'on',
+          'off',
+          // web & native
+          'none', // used instead of 'off'
+          'sentences', // used instead of 'on'
+          'words',
+          'characters'
```
necolas: This test is just for native, and if we're supporting all 6 values it's confusing (and inaccurate) to have comments like "web only".

wlabu:
On an aside - I had a look at the failing CI jobs before force-pushing:
- https://github.com/facebook/react-strict-dom/actions/runs/20122864676

Wondering if it might be related to this: https://github.com/peter-evans/create-or-update-comment/issues/395

wlabu:
Sorry to comment on a closed PR - I'm not even sure you'll be notified.
All I really wanted to ask is whether you have an idea of when the next release will be cut?

I'd love to use this change downstream in a project :)

necolas:
Sorry for the delay. Just published 0.0.55

wlabu:
Thank you!

wlabu:
👋 Once again, sorry for commenting on a closed PR. I didn't want to create a fresh issue or discussion just for this.

When trying to upgrade RSD to 0.0.55 (we use yarn v4, so `yarn up 'react-strict-dom'`), we get an error:

```sh
› yarn up 'react-strict-dom'

➤ YN0000: · Yarn 4.9.2
➤ YN0000: ┌ Resolution step
➤ YN0082: │ postcss-react-strict-dom@npm:0.0.55: No candidates found
➤ YN0000: └ Completed
➤ YN0000: · Failed with errors in 0s 120ms
```

I've checked main, and the `postcss` package was bumped.

However, I don't see a 0.0.55 [Release](https://github.com/facebook/react-strict-dom/releases), and the latest version [on npmjs.com](https://www.npmjs.com/package/postcss-react-strict-dom?activeTab=versions) is 0.0.54

Could I kindly ask you to have a look?

necolas:
Should be published now

PR #444, opened 2026-01-14
add react-strict-animated example app

vincentriemer:
add react-strict-animated example app

This diff adds an example app showing off how to use the react-strict-animated library in an expo app. Currently when the app renders on the web there are some animation issues but the API calls are correct and I'll be addressing the composite animation issues in a later diff.

Web Rendering:

https://github.com/user-attachments/assets/8680640d-adf5-4ca3-a289-7f270c0f4ac6

iOS Rendering:

https://github.com/user-attachments/assets/85519b2b-9007-4470-8875-b51950d5afa7


MoOx:
@martinbooth this PR should be considered for merging

PR #447, opened 2026-01-28
fix: remove fontScale from CSS relative units to prevent double scaling

axyz:
By default react-native textual elements have the `allowFontScaling` prop set to true, effectively making `px` values to behave like `rem`/`em` values. The existing polyfill, though, scales the px values using the detected `fontScale`. This leads to a double amount of scaling on the native app.

This PR remove the fontScale multiplier, leaving the scaling to the native capability, while keeping the polyfill logic to translate em/rem to appropriate px values.

### DIFF
--- a/packages/react-strict-dom/tests/css/__snapshots__/css-create-test.native.js.snap
+++ b/packages/react-strict-dom/tests/css/__snapshots__/css-create-test.native.js.snap
@@ -346,13 +346,6 @@ exports[`css.create() properties fontSize 1`] = `
 }
 `;
 
-exports[`css.create() properties fontSize scaling fontScale:2 1`] = `
-{
-  "boxSizing": "content-box",
-  "fontSize": 80,
-  "position": "static",
-}
-`;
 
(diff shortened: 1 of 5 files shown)


packages/react-strict-dom/tests/css/__snapshots__/css-create-test.native.js.snap
```diff
@@ -2077,35 +2070,35 @@ exports[`css.create() values: length units 10 "rem" units are resolved to pixels
 exports[`css.create() values: length units 10 "vh" units are resolved to pixels: vh 1`] = `
 {
   "boxSizing": "content-box",
-  "height": NaN,
+  "height": -100,
   "position": "static",
-  "width": NaN,
+  "width": 100,
```
axyz, line 2075: it seems that removing the `fontScale` accidentally fixed this preexisting bug where the resolved value was NAN, I'm assuming it could have been a lack of mocking in the test though?

PR #451, opened 2026-02-10
Add default "overflow: visible" to Text

NickGerleman:
iOS and Android both historically clip everything outside the bounds of the paragraph/top-level text element. This includes clipping content, when `line-height` would set the bounds to be smaller than the text to be drawn. This is equivalent to setting `overflow: 'hidden'` manually on a paragraph.

Changing this default behavior in React Native is scary and breaking and not something we are taking on right now, but with Facsimile/PreparedLayoutTextView on Android, it can be controlled with opt-in `overflow: visible`, for web style clipping.

This change defaults to `overflow: 'visible'` to get that web style behavior by default (e.g. no clipping the bounds of `TextView`/`<p>` when `line-height` makes text larger than the bounds.


NickGerleman:
@mellyeliu this should prevent existing line height clipping issue on new Android text stack rolling out.

PR #453, opened 2026-02-13
Add babel codemod for RN to RSD transform

necolas:
Transforms to convert existing RN code to RSD code


edkimmel:
What's the state of this, and what is the estimated impact in runtime overhead for the translation layer from RSD to RN?  The current overhead is impactful enough that we have been maintaining a separate RN fork for all hot flow views, requiring any style props to be `{ native: StyleProp, web: StyleX }` to avoid any style translations.

PR #457, opened 2026-02-26
fix/456-default-value-ignored-dark-mode

### LINKED ISSUE #456: Default value is ignored if device color scheme is dark
### Describe the issue

When using dark mode on device the default value of `defineVars` with media-queries other than `preferes-color-scheme` are ignored. If using light mode the default value is used as expected.

From researching this a bit if think the issue is at line 71 in `customProperties.js`:
```javascript
   if (variableValue != null) {
    if (typeof variableValue === 'object' && variableValue.default != null) {
      let defaultValue = variableValue.default;
      if (colorScheme === 'dark') {
        defaultValue = variableValue['@media (prefers-color-scheme: dark)'];
      }
      return defaultValue;
    }
    return variableValue;
```
It looks like the if-check `if (colorScheme === 'dark')` should also check if the `variableValue['@media (prefers-color-scheme: dark)']` exists before setting the defaultValue.

### Expected behavior

When device is set to dark mode styling should behave the same as in light mode.

### Steps to reproduce

1. Define a css-var like this:
```javascript
// fontSize.css.js
export const fontSize = css.defineVars({
  small: {
    default: '32px',
    '@media (min-width: 800px)': '20px',
  }
})
```
2. Use the var in a component:
```javascript
import { fontSize } from './fontSize.css';

export function ComponentA() {
	return <html.p style={styles.componentA}>Responsive test</html.p>;
}

const styles = css.create({
	componentA: {
		fontSize: fontSize.small,
	},
});
```
3. Check the component on device/simulator in dark-mode.

### Test case

_No response_

### Additional comments

Images from the above example running in our test-app. (Ignore the ugly theme switcher in the upper right)

Light-mode:
Image

Dark-mode:
Image

orhels:
fixes #456 

* Verify that variableValue['@media (prefers-color-scheme: dark)'] is defined before using it as defaultValue
* Added test for css.defineVars

This fix aligns the behavior across light and dark mode.  A better solution would probably be to actually support a subset of the media query specification, but I think that would require some more effort and planing. 

The tests are a little strange as they tests that using media-queries other than `@media (prefers-color-scheme: dark)` will return the default value, and not that other media queries actually works. I used the `css-create-queries-test.native.js` as a template and tried to keep the test names a concise as possible.

### DIFF
--- a/packages/react-strict-dom/src/native/css/customProperties.js
+++ b/packages/react-strict-dom/src/native/css/customProperties.js
@@ -68,7 +68,10 @@ function resolveVariableReferenceValue(
   if (variableValue != null) {
     if (typeof variableValue === 'object' && variableValue.default != null) {
       let defaultValue = variableValue.default;
-      if (colorScheme === 'dark') {
+      if (
+        colorScheme === 'dark' &&
+        variableValue['@media (prefers-color-scheme: dark)'] != null
+      ) {
         defaultValue = variableValue['@media (prefers-color-scheme: dark)'];
       }
       return defaultValue;
--- a/packages/react-strict-dom/tests/css/css-define-vars-test.native.js
+++ b/packages/react-strict-dom/tests/css/css-define-vars-test.native.js
@@ -1,0 +1,105 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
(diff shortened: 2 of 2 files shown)


efoken:
@martinbooth can you have a look and merge this please?

PR #464, opened 2026-03-18
Update hermes-parser related dependencies

gkz:
Update hermes-parser related dependencies to the latest


gkz:
The two failing benchmarks signals do not look related to this change. They are also failing on other PRs

PR #475, opened 2026-05-12
Adopt React Native 0.82 DOM Node APIs

yaminyassin:
## Summary

Bumps `react-native` peer dep to `>=0.82.0` and adopts the stable DOM Node APIs from RN 0.82. The native ref polyfill becomes an overlay on top of RN's host node instead of a parallel object that re-implements DOM traversal.

The first commit on this branch (4a49d1f) wrapped the node in a `Proxy`. After review feedback that Hermes can't inline through Proxy traps, follow-up 8fbae4f replaces it with `Object.create(node)` prototype delegation: strict-dom defines its overrides as own properties on the wrapper, and everything else delegates to the node via the prototype chain. Same observable behavior, static hidden class Hermes can optimize, no per-access bound-function allocations.

Tracks #462.

## What's still polyfilled

- `nodeName`: uppercase DOM name. RN exposes `tagName` as `'RN:View'`.
- `getBoundingClientRect` and the length getters (`offsetWidth`, `clientHeight`, ...): divided by the active `viewportScale`. Only installed when scale isn't 1, so the common path skips them.
- `.complete`: returns `false` when the underlying Image node doesn't expose it.
- `<input>` / `<textarea>` `setSelectionRange` / `selectionStart` / `selectionEnd`: polyfilled over `setSelection`. Each one is gated on the underlying property being missing, so the polyfill switches itself off the day RN exposes the W3C API.

## What forwards to the RN node

`ownerDocument`, `getRootNode`, `parentNode`, `parentElement`, `children`, `childNodes`, sibling navigation, `contains`, `compareDocumentPosition`, pointer-capture, legacy `measure*` and `setNativeProps`. With the node as the wrapper's prototype, RN's prototype methods invoked on the wrapper see `this === wrapper`; the symbol-keyed internals they consult (`INSTANCE_HANDLE_KEY`, `OWNER_DOCUMENT_KEY` from `NodeInternals.js`) are reachable through the chain back to the node, so RN's implementations resolve their backing state correctly.

`instanceof` keeps working. The chain is `wrapper → node → ReactNativeElement.prototype → ReadOnlyElement.prototype → ReadOnlyNode.prototype`, one link longer than the pre-PR clone but the class prototypes are still on it. Verified on a live iPad simulator (`wrapper.constructor.name === 'ReactNativeElement'`) and in Chrome (`'HTMLDivElement'`).

Identity caching via `WeakMap<Node, Node>` is preserved, so the same RN node yields the same wrapped ref across renders.

## Descriptor flags

`configurable: true` is set on every override so `Object.defineProperty` can redefine them later (tests, devtools, or a future fix to viewport-scale memoization). Value descriptors (`nodeName`, `getBoundingClientRect`, `setSelectionRange`) are not `writable`, matching the DOM spec for read-only properties. Strict-mode assignments like `wrapper.setSelectionRange = ...` now throw `TypeError` instead of silently shadowing. Demoed live in the example apps.

## Breaking

Peer dep `react-native >=0.79.5` → `>=0.82.0`. Apps on older RN stay on the previous react-strict-dom release.

## Notes for reviewers

- **Hermes / Proxy concern.** Addressed in 8fbae4f. The delegation variant has a 1-to-15 own-property hidden class (size depends on viewport scale and tag), no traps, no per-access binding. The CSS benchmarks in `benchmarks/` don't exercise the ref read path, so the existing numbers aren't informative for this change. Happy to add a targeted micro-benchmark if useful.

- **Expo SDK.** No SDK pairs exactly with RN 0.82 (SDK 53 = 0.79, SDK 54 ≈ 0.81, SDK 55 = 0.83). The example apps move to SDK 55 / RN 0.83.6, which satisfies the new peer dep and ships the same DOM Node surface.

- **React devDeps.** RN 0.82+ requires React `>=19.1.1`. The workspace root and library packages' devDependencies pin `react` / `react-dom` / `react-test-renderer` to `~19.2.0` to keep a single React instance across the monorepo.

- **Flow types in 0.83.** `Animated.createAnimatedComponent` is now a single-type-arg generic; fixed by dropping the over-specified two-type-arg form. `ImageProps` / `TextInputProps` became exact; the wide spread of strict-dom's `ReactNativeProps` is covered by two targeted `$FlowFixMe` annotations. Narrowing that spread properly is a separate follow-up.

- **Ten ref tests** in `tests/html/html-refs-test.native.js` pin every wrapped-ref behavior: uppercase nodeName, gBCR pass-through at scale 1 and scaled at non-1, DOM Node API pass-through (`ownerDocument`, `getRootNode`, `childNodes`, `children`), strict-ref identity stability, `.complete` fallback and pass-through, and the input selection trio. All ten pass against both the original Proxy commit and the prototype-delegation follow-up.

- **One pre-existing latent issue noted.** `memoizedStrictRefs` is keyed only on `node`, not on `(node, viewportScale)`. A node rendered first outside `ViewportProvider` and then inside (or with a changing scale) keeps the wrapper from the first wrap. Predates this PR; calling out for a follow-up. Not blocking. I worked around it locally in a realtime demo by putting `key={viewportWidth}` on the inner reporter to force a remount.

### DIFF
--- a/apps/expo-app/app.json
+++ b/apps/expo-app/app.json
@@ -12,7 +12,9 @@
       "resizeMode": "contain",
       "backgroundColor": "#ffffff"
     },
-    "assetBundlePatterns": ["**/*"],
+    "assetBundlePatterns": [
+      "**/*"
+    ],
     "android": {
       "adaptiveIcon": {
         "foregroundImage": "./assets/adaptive-icon.png",
@@ -25,6 +27,9 @@
     "web": {
       "bundler": "metro",
       "favicon": "./assets/favicon.png"
-    }
+    },
+    "plugins": [
+      "expo-build-properties"
+    ]
   }
 }
--- a/apps/expo-app/package.json
+++ b/apps/expo-app/package.json
@@ -10,15 +10,15 @@
     "dev:web": "expo start --web --clear"
   },
   "dependencies": {
-    "@expo/metro-runtime": "~5.0.4",
+    "@expo/metro-runtime": "~55.0.11",
     "example-ui": "0.0.55",
-    "expo": "^53.0.11",
-    "expo-build-properties": "~0.14.6",
-    "expo-status-bar": "~2.2.3",
-    "react": "~19.0.0",
-    "react-dom": "~19.0.0",
-    "react-native": "~0.79.5",
-    "react-native-web": "~0.20.0",
+    "expo": "^55.0.0",
+    "expo-build-properties": "~55.0.13",
+    "expo-status-bar": "~55.0.6",
+    "react": "19.2.0",
+    "react-dom": "19.2.0",
+    "react-native": "0.83.6",
+    "react-native-web": "^0.21.0",
     "react-strict-dom": "0.0.55"
   },
   "devDependencies": {
--- a/apps/platform-tests/package.json
+++ b/apps/platform-tests/package.json
@@ -10,14 +10,14 @@
     "dev:web": "expo start --web --clear"
   },
   "dependencies": {
-    "@expo/metro-runtime": "~5.0.4",
-    "expo": "^53.0.11",
-    "expo-build-properties": "~0.14.6",
-    "expo-status-bar": "~2.2.3",
-    "react": "~19.0.0",
-    "react-dom": "~19.0.0",
-    "react-native": "~0.79.5",
-    "react-native-web": "~0.20.0",
+    "@expo/metro-runtime": "~55.0.11",
+    "expo": "^55.0.0",
+    "expo-build-properties": "~55.0.13",
+    "expo-status-bar": "~55.0.6",
+    "react": "19.2.0",
+    "react-dom": "19.2.0",
+    "react-native": "0.83.6",
+    "react-native-web": "^0.21.0",
     "react-strict-dom": "0.0.55"
   },
   "devDependencies": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -44,9 +44,9 @@
         "patch-package": "^8.0.0",
         "prettier": "^3.3.3",
         "prettier-plugin-hermes-parser": "0.36.1",
-        "react": "~19.2.0",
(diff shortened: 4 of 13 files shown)


yaminyassin:
Related follow-up: #477: switches strict-dom from the deep import `react-native/Libraries/Text/TextAncestor` to the public `unstable_TextAncestorContext` API exposed in [facebook/react-native#52368](https://github.com/facebook/react-native/pull/52368). Logically depends on this PR (the bumped peer dep is what makes the public API guaranteed available). Once both land, the corresponding RN-side cleanup can delete [the compatibility shim](https://github.com/facebook/react-native/blob/main/packages/react-native/Libraries/Text/TextAncestor.js).

martinbooth:
Hey, thanks for this! It looks good, however can we use the object clone approach rather than the Proxy? I know that using a Proxy here is much more ergonomic than the way the code is currently written, but the performance of using Proxy is substantially slower than the existing code. Due to the very dynamic nature of a Proxy hermes (the js runtime most react native applications use) is unable to apply the optimization static code can benefit from

yaminyassin:
Restored the original approach and improved a couple of things!
Tell me what you think. i've already validated and profiled the behaviors on my devices and everything looks good 😊

yaminyassin:
it looks like CI i failing because of auth issues, @martinbooth is this normal?

martinbooth:
I think if you rebase on latest main it should work (although that will also be my test to see if https://github.com/facebook/react-strict-dom/pull/481 fixes the issue).

The issue was with uploading a comment to the PR reporting size and perf data (if it runs outside of facebook's repo, you only get a readonly auth token and this fails).. I've split the steps so this shouldn't be an issue any more.. at least that's the plan

yaminyassin:
Thanks Martin! Rebased onto latest main.  Could you kick off the run when you get a sec? That'll be the real test of whether splitting the steps fixed the token issue. 🙏

PR #477, opened 2026-05-13
Use public unstable_TextAncestorContext API

yaminyassin:
React Native 0.81 promoted `TextAncestorContext` to a public API in [facebook/react-native#52368](https://github.com/facebook/react-native/pull/52368), so we no longer need to reach into
  `react-native/Libraries/Text/TextAncestor`.

  The deep-import file on the RN side [only exists as a shim for this repo](https://github.com/facebook/react-native/blob/main/packages/react-native/Libraries/Text/TextAncestor.js):

  ```js
  // Compatibility module for ReactStrictDOMTextAncestorContext.native.js.flow (react-strict-dom)
  // TODO(huntie): Delete after we've fixed this cross-repo reference

  import TextAncestorContext from './TextAncestorContext';
  export default TextAncestorContext;
  ```

  So merging this clears the cross-repo reference the TODO is waiting on, and the shim can be deleted on the RN side afterwards (I'll open that PR once this lands).

  It's the same Context object under the hood — no behavior change, snapshots are unchanged.

  ### Changes
  - `src/native/react-native/TextAncestorContext.js`: import `unstable_TextAncestorContext` from `react-native` instead of the deep path.
  - `tests/__mocks__/react-native/index.js`: export `unstable_TextAncestorContext` so the mock matches.
  - `tests/__mocks__/react-native/Libraries/Text/TextAncestor.js`: deleted, no longer referenced.

  ### Depends on
  #475 — `unstable_TextAncestorContext` needs React Native ≥ 0.81, which the peer-dep bump there handles.


yaminyassin:
Cross-repo follow-up: [facebook/react-native#56813](https://github.com/facebook/react-native/pull/56813) (draft) deletes the now-unneeded compatibility shim on the React Native side. It's gated on this PR landing and a strict-dom release shipping with the new import path.

martinbooth:
Thank you! I think the branch associated with this pull request has several other commits/features though. If you're able to separate the [Use public unstable_TextAncestorContext API](https://github.com/facebook/react-strict-dom/pull/477/changes/1f844c0e0cbd739556ca43f894826a286cd1bb36) commit from the other changes I will merge

yaminyassin:
Done! only the [Use public unstable_TextAncestorContext API](https://github.com/facebook/react-strict-dom/pull/477/changes/1f844c0e0cbd739556ca43f894826a286cd1bb36) commit remains. 

Quick heads-up before you merge: unstable_TextAncestorContext needs RN ≥ 0.81, but main's peer dep is still >=0.79.5. So on its own this could break for users on 0.79.x or 0.80.x.

Easiest fix is probably merging #475 first since it bumps the peer dep anyway. But if you'd rather take this one first, I can tack on a >=0.81.0 bump here. Let me know what you prefer.

yaminyassin:
@martinbooth can you have a look? if this merges we'll be able to remove the shim in RN [facebook/react-native#56813](https://github.com/facebook/react-native/pull/56813)

PR #484, opened 2026-05-24
Ignore undefined style values on native

### LINKED ISSUE #452: styleq: <cssAtom> typeof undefined is not "string" or "null".
### Describe the issue

I recently updated from RSD 0.33 to 0.55 and started encountering this error.

Returning undefined for a property in an inline style function now causes this error.  I believe these values were previously ignored / stripped by RSD.

### Expected behavior

If this is an intentional change, I would expect the typescript types for this function to disallow undefined values.

### Steps to reproduce

```
const styles = css.create({
  gap: (gap?: number) => ({
    rowGap: gap, // An undefined gap will result in styleq: rowGap-abcdef typeof undefined is not "string" or "null".
  }),
})
```

### Test case

_No response_

### Additional comments

_No response_

adityasingh2400:
Fixes #452

On native, returning `undefined` for a property from an inline style function (or any style object with an undefined value) leaks that key through `processStyle` and into the style resolver, which throws `styleq: <cssAtom> typeof undefined is not "string" or "null"`. This is a regression; these values used to be dropped, and on web undefined values are simply ignored.

The loop in `processStyle` branches on `object`/`string`/`number`, so an undefined value falls through to the catch-all assignment and ends up in the result as `{ rowGap: undefined }`. This adds an early `continue` for undefined values so the property is omitted, matching web behavior and letting inline style functions return undefined to opt out of a property.

`null` is left untouched since the resolver accepts it.

Testing: added a regression test under `css.create() > values: general` that asserts the property is absent when the inline function returns undefined. Confirmed it fails on `main` and passes with the fix. Full native + web jest suites pass (888 tests, 711 snapshots unchanged), Flow and ESLint clean.

### DIFF
--- a/packages/react-strict-dom/src/native/css/processStyle.js
+++ b/packages/react-strict-dom/src/native/css/processStyle.js
@@ -73,6 +73,14 @@ export function processStyle(
   for (const propName in style) {
     const styleValue = style[propName];
 
+    // Drop "undefined" values so they don't reach the underlying style
+    // resolver, which only accepts strings, numbers, objects or null. This
+    // mirrors web behavior, where undefined values are ignored, and lets
+    // inline style functions return undefined to opt out of a property.
+    if (styleValue === undefined) {
+      continue;
+    }
+
     if (skipValidation !== true && !isAllowedStyleKey(propName)) {
       if (__DEV__) {
         warnMsg(`unsupported style property "${propName}"`);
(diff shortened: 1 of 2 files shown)


MoOx:
Any reason you closed this ?

adityasingh2400:
Sorry @MoOx, no reason related to the change itself. I closed it by accident while bulk cleaning up stale forks on my profile, and since the fork got deleted I could not reopen this one directly. I have restored the exact same branch and opened #498 to continue it. Happy to address review feedback there.

PR #485, opened 2026-05-24
[react-strict-animated] Fix incorrect error message for invalid mass in SpringAnimation

adityasingh2400:
## Summary

The `mass` validation in the `SpringAnimation` constructor throws the wrong error message. It reports a damping problem when an invalid mass is supplied, because the message was copy-pasted from the preceding damping check.

```js
if (this.#damping <= 0) {
  throw new Error('Damping value must be greater than 0');
}
if (this.#mass <= 0) {
  throw new Error('Damping value must be greater than 0'); // should be Mass
}
```

A user passing `mass: 0` (or a negative mass) gets told their damping value is wrong, which points them at the wrong config field. React Native's equivalent `SpringAnimation` throws `'Mass value must be greater than 0'` here.

## Fix

Use the correct message for the mass check.

## Test

Added `SpringAnimation-test.js` covering all three constructor validations (stiffness, damping, mass). The mass case fails on `main` (asserts `'Mass value must be greater than 0'`) and passes with the fix.

`flow check` and `eslint` are clean.

### DIFF
--- a/packages/react-strict-animated/src/web/__tests__/SpringAnimation-test.js
+++ b/packages/react-strict-animated/src/web/__tests__/SpringAnimation-test.js
@@ -1,0 +1,30 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow strict-local
+ */
+
+import SpringAnimation from '../animations/SpringAnimation';
+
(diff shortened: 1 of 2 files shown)


MoOx:
@adityasingh2400 Care to comment about closing something that could help ?

adityasingh2400:
Sorry @MoOx, same story as #484, this was an accidental close during a bulk fork cleanup, not a decision about the fix. The fork was deleted so I could not reopen it, so I restored the branch and opened #499 to carry it forward. Glad to iterate on it there.

PR #492, opened 2026-06-01
Fix html.button staying unclickable on Android after toggling disabled back to false (#491)

### LINKED ISSUE #491: Button remains unclickable after disabled state changes
### Describe the issue

On Android if `html.button` state is set to disabled and then back to enabled the pressable itself remains unclickable, while it's children are clickable as expected. As with other similar issues in React Native a workaround is to set a `key` based on disabled state, but this issue doesn't seem to come from React Native, as it doesn't happen in React Native.

On iOS the button regains clickability as expected.

### Expected behavior

Like on iOS, the pressable on Android should become clickable when the disabled state changes back to `false`.

### Steps to reproduce

Platform: Android
OS version: 16.0 ("Baklava"); API 36.0
React version: 19.2.3
React Native version: 0.85.3
RSD version: 0.0.55

1. Use the below example:

```tsx
import { useState } from "react";
import { Button, Pressable, Text, View } from "react-native";
import { css, html } from "react-strict-dom";

export default function RootLayout() {
  const [buttonDisabled, setButtonDisabled] = useState(false);
  const [rnPresses, setRnPresses] = useState(0);
  const [rsdPresses, setRsdPresses] = useState(0);

  return (
    <View
      style={{ flex: 1, backgroundColor: "white", alignItems: "center", justifyContent: "center" }}
    >
      <Button
        title={"Toggle button disabled - now: " + buttonDisabled.toString()}
        onPress={() => {
          setButtonDisabled((prev) => !prev);
        }}
      />
      <View>
        <Text>RN Pressable - {rnPresses} presses</Text>
        <Pressable
          disabled={buttonDisabled}
          style={({ pressed }) => [
            { backgroundColor: "lightgreen", padding: 16 },
            pressed && { opacity: 0.5 },
          ]}
          onPress={() => {
            setRnPresses((prev) => prev + 1);
          }}
        >
          <Text style={{ backgroundColor: "black", color: "white", padding: 8 }}>Press me</Text>
        </Pressable>
      </View>
      <View>
        <Text>RSD Button - {rsdPresses} presses</Text>
        <html.button
          disabled={buttonDisabled}
          style={styles.button}
          onClick={() => {
            setRsdPresses((prev) => prev + 1);
          }}
        >
          <html.span style={styles.buttonText}>Press me</html.span>
        </html.button>
      </View>
    </View>
  );
}

const styles = css.create({
  button: {
    backgroundColor: "lightblue",
    opacity: {
      default: 1,
      ":active": 0.5,
    },
    padding: 16,
  },
  buttonText: {
    backgroundColor: "black",
    color: "white",
    padding: 8,
  },
});
```

2. Run the app on Android device/emulator
3. Verify that, for both RN and RSD cases, both pressable and it's children trigger press event (press count increases)
4. Toggle button disabled state to true
5. Toggle button disabled state to false
6. Verify that for RN case both pressable and children regained clickability, while in RSD case only the label can be clicked to trigger an event

### Test case

_No response_

### Additional comments

_No response_

MoOx:
Closes #491

## Problem

On Android, when an `html.button` (or any RSD element rendered as a `Pressable`) has its `disabled` state set to `true` and then back to `false`, the pressable's own surface remains unclickable — only its children continue to trigger the press handler. On iOS the button recovers correctly.

This does **not** reproduce with a plain React Native `<Pressable disabled={...} />`, and the usual `key`-based remount workaround "fixes" it — both signs that the regression is in how RSD forwards the prop, not in React Native itself.

## Root cause

In `createStrictDOMComponent.js`, the `disabled` prop was forwarded to the underlying `Pressable` **only when it was `true`**, and never reset when it became `false` (`disabled` is intentionally not passed through `useNativeProps`):

```js
if (NativeComponent === ReactNative.Pressable) {
  if (props.disabled === true) {
    nativeProps.disabled = true;
    nativeProps.focusable = false;
  }
}
```

So `Pressable` saw `disabled` toggle `true → undefined`, never `true → false`. React Native's `Pressable` only merges `disabled` into its `accessibilityState` when it's non-null:

```js
// react-native/Libraries/Components/Pressable/Pressable.js
_accessibilityState =
  disabled != null ? {..._accessibilityState, disabled} : _accessibilityState;
```

With `undefined`, `disabled` drops out, so `accessibilityState.disabled` goes `true → undefined` instead of `true → false`. On Android, `accessibilityState.disabled = true` sets the native view to `enabled = false`; reverting to `undefined` (rather than an explicit `false`) does not re-enable it. A disabled Android `ViewGroup` still dispatches touches to its children (so the inner text kept working) but won't claim touches on its own surface — exactly the reported symptom. iOS has no equivalent native "enabled" latch, so it recovered regardless.

Plain RN `Pressable` works because the user passes an explicit boolean that always resets to `false`.

## Fix

Always forward `disabled` to the `Pressable` as an explicit boolean, so it transitions `true ↔ false` and Android correctly resets the native enabled state:

```js
if (NativeComponent === ReactNative.Pressable) {
  // Always pass an explicit boolean so that toggling 'disabled' resets the
  // native view's enabled state. Passing 'undefined' when re-enabling leaves
  // 'accessibilityState.disabled' unset, which on Android keeps the view's
  // native 'enabled' state false and makes the Pressable unclickable.
  nativeProps.disabled = props.disabled === true;
  if (props.disabled === true) {
    nativeProps.focusable = false;
  }
}
```

## Note on `focusable` (intentionally left conditional)

`focusable` does **not** need the same symmetric treatment, for two reasons:

1. **It never reaches native as `undefined`.** `Pressable` normalizes it to an explicit boolean on every render via `focusable: focusable !== false`. So when RSD leaves it unset on re-enable, the native View receives `focusable={true}`; when disabled it receives `focusable={false}`. The value always toggles `false ↔ true` at the native layer, so there's no stale-state problem like `disabled` had.

2. **Making it unconditional would regress `tabIndex`.** `focusable` is also derived from `tabIndex` in `useNativeProps` (`nativeProps.focusable = !tabIndex`). The current code only forces `focusable = false` *while disabled* and otherwise preserves the `tabIndex`-derived value. Changing it to something like `focusable = props.disabled !== true` would clobber `tabIndex` (e.g. an enabled button with `tabIndex={-1}` would wrongly become focusable).

So the `disabled` fix alone is sufficient, and the asymmetric `focusable` handling is correct and intentional.

## Tests

- Added a `<button>` regression test asserting that after toggling `disabled` from `true` to `false`, the rendered `Pressable` emits an explicit `disabled={false}` (rather than dropping the prop).
- Full native test suite passes; Flow is clean.

### Alternative: fixing this without emitting `disabled={false}` on every `Pressable`

The current fix always forwards an explicit boolean:

```js
nativeProps.disabled = props.disabled === true;
```

This is the most defensive option — it even survives a consumer toggling `disabled` between `true` and `undefined` — but it has a cost: **every** element that renders as a `Pressable` (every `html.button`, plus any element upgraded to a `Pressable` because it has an event handler) now carries a `disabled={false}` prop. That's why ~20 native snapshots changed even though their behaviour didn't.

#### Preferred alternative — only forward `disabled` when it is explicitly provided

```js
if (NativeComponent === ReactNative.Pressable) {
  if (props.disabled != null) {
    // forward the consumer's value as-is (true *or* false)
    nativeProps.disabled = props.disabled;
  }
  if (props.disabled === true) {
    nativeProps.focusable = false;
  }
}
```

Why this avoids the snapshot churn:

- An element that **never** sets `disabled` (the vast majority — the default button render, event-handler views, etc.) emits no `disabled` prop at all, exactly as before → **those snapshots stay unchanged**.
- The moment a consumer controls `disabled` with state (`disabled={buttonDisabled}`, as in the repro), the value is a real boolean on **both** transitions, so the `Pressable` receives `disabled={true}` ↔ `disabled={false}`. That's precisely what resets Android's native `enabled` state and fixes the bug.

So the only snapshots that would change are ones that explicitly pass `disabled={false}` (there are essentially none today), instead of all of them.

#### Trade-off to be aware of

This version relies on the consumer passing an explicit boolean in both directions — which is the idiomatic controlled pattern and matches HTML's boolean-attribute semantics. It would **not** rescue an unusual pattern like `disabled={cond ? true : undefined}`, because then `disabled` is `null`/absent on the "enabled" render and we'd skip forwarding it (re-introducing the `true → undefined` transition that Android doesn't reset). The `=== true` version in this PR is immune to that at the price of the snapshot noise.

If we'd rather keep the snapshots clean, I'm happy to switch to the `!= null` form — it covers the reported case and every realistic controlled usage.

#### Root cause is arguably upstream

The underlying issue is that on Android, React Native does not reset the native view's `enabled` flag when `accessibilityState.disabled` goes `true → undefined` (only `true → false` resets it). A defensive fix in RSD makes sense regardless, but it may be worth reporting/fixing in React Native so `undefined` and `false` behave consistently across platforms.

### DIFF
--- a/packages/react-strict-dom/src/native/modules/createStrictDOMComponent.js
+++ b/packages/react-strict-dom/src/native/modules/createStrictDOMComponent.js
@@ -89,12 +89,23 @@ export function createStrictDOMComponent<T, P extends StrictProps>(
 
     // Component-specific props
 
+    // Forward `disabled` to the Pressable only when it is an explicit boolean,
+    // so that toggling `true` <-> `false` resets the native view's enabled
+    // state. On Android, re-enabling by dropping the prop (`true` -> `undefined`)
+    // leaves `accessibilityState.disabled` unset and the native `enabled` state
+    // stays false, making the Pressable unclickable. Emitting only for explicit
+    // booleans keeps elements that never opt into `disabled` unchanged; the
+    // `true` -> `undefined` transition is intentionally not covered, but a
+    // controlled `disabled` toggles `true` <-> `false`.
     if (NativeComponent === ReactNative.Pressable) {
       if (props.disabled === true) {
         // $FlowFixMe[react-rule-hook-mutation]
         nativeProps.disabled = true;
         // $FlowFixMe[react-rule-hook-mutation]
         nativeProps.focusable = false;
+      } else if (props.disabled === false) {
+        // $FlowFixMe[react-rule-hook-mutation]
+        nativeProps.disabled = false;
       }
     }
 
--- a/packages/react-strict-dom/tests/html/__snapshots__/html-test.native.js.snap
+++ b/packages/react-strict-dom/tests/html/__snapshots__/html-test.native.js.snap
@@ -216,6 +216,22 @@ exports[`<html.*> (native polyfills) polyfills: layout default flex layout: flex
 </ViewNativeComponent>
 `;
 
+exports[`<html.*> (native polyfills) polyfills: props <button> "disabled" prop 1`] = `
+<Pressable
+  disabled={true}
+  focusable={false}
+  ref={[Function]}
+  role="button"
+  style={
+    {
+      "borderWidth": 1,
+      "boxSizing": "content-box",
+      "position": "static",
+    }
+  }
+/>
+`;
+
 exports[`<html.*> (native polyfills) polyfills: props <img> "src" prop with loading props 1`] = `
 <Image
   crossOrigin="use-credentials"
--- a/packages/react-strict-dom/tests/html/html-test.native.js
+++ b/packages/react-strict-dom/tests/html/html-test.native.js
@@ -500,6 +500,31 @@ describe('<html.*> (native polyfills)', () => {
         expect(root.toJSON()).toMatchSnapshot();
       });
     });
+
+    describe('<button>', () => {
+      test('"disabled" prop', () => {
+        let root;
+        act(() => {
+          root = create(<html.button disabled={true} />);
+        });
+        expect(root.toJSON()).toMatchSnapshot();
+      });
+
+      // Re-enabling must emit an explicit "disabled={false}" rather than
+      // dropping the prop. On Android, leaving "disabled" unset keeps the
+      // native view's enabled state false and the button stays unclickable.
+      test('"disabled" prop resets to an explicit false when re-enabled', () => {
+        let root;
+        act(() => {
+          root = create(<html.button disabled={true} />);
+        });
+        expect(root.toJSON().props.disabled).toBe(true);
+        act(() => {
+          root.update(<html.button disabled={false} />);
+        });
+        expect(root.toJSON().props.disabled).toBe(false);
+      });
+    });
   });
 
   describe('polyfills: inheritence', () => {


packages/react-strict-dom/src/native/modules/createStrictDOMComponent.js
```diff
@@ -90,9 +90,13 @@ export function createStrictDOMComponent<T, P extends StrictProps>(
     // Component-specific props
 
     if (NativeComponent === ReactNative.Pressable) {
+      // Always pass an explicit boolean so that toggling 'disabled' resets the
+      // native view's enabled state. Passing 'undefined' when re-enabling leaves
+      // 'accessibilityState.disabled' unset, which on Android keeps the view's
+      // native 'enabled' state false and makes the Pressable unclickable.
+      // $FlowFixMe[react-rule-hook-mutation]
+      nativeProps.disabled = props.disabled === true;
```
pmleczek: I'm wondering if it wouldn't be better to implement it as:

```ts
if (props.disabled === true) {
  nativeProps.disabled = true;
  ...
} else if (props.disabled === false) {
  nativeProps.disabled = false;
}
```

Because `disabled` prop can be either:
- `false`
- `true`
- `undefined`

And explicitly passing boolean value (false) in case it's `undefined` seems to affect e.g. accessibility state of the component (you can check it e.g. by viewing test snapshots). So as this issue only seems to happen when `disabled` value changes, it'd probably be better to narrow fix down to those cases and to leave undefined state as it was 🤔

MoOx:
Good point — I've updated the PR to narrow it to the explicit-boolean cases:

```js
if (props.disabled === true) {
  nativeProps.disabled = true;
  nativeProps.focusable = false;
} else if (props.disabled === false) {
  nativeProps.disabled = false;
}
```

This fixes the reported issue (a controlled `disabled` toggles `true ↔ false`, and both go through explicit branches that reset the native state) while avoiding a behavioural change for elements that never opt into `disabled`. The accessibility argument is the clincher: forcing `accessibilityState.disabled = false` on every `Pressable` (including plain views with an `onClick`) changes their a11y semantics unnecessarily — and that's exactly what was showing up as snapshot churn. Keeping `undefined` untouched is the smaller blast radius and the more faithful polyfill behaviour, and no existing snapshots change.

The one case this consciously doesn't cover is toggling `true → undefined` directly (e.g. `disabled={cond || undefined}` or a conditional spread), which would re-introduce the original `true → undefined` transition. That's a non-idiomatic pattern for a controlled `disabled` (which is almost always bound to a boolean), so it seems like a fine trade-off — I added a code comment documenting it so it's an explicit decision rather than an accidental gap.

For context on the RN-version sensitivity you noticed (0.83 OK, 0.85 KO): the RSD code is identical across RN versions, so the prop sequence we send (`disabled: true → undefined`) doesn't depend on the version — the latent asymmetry has always been there. What changed between 0.83 and 0.85 is purely RN's **Android native** side: whether the view's native `enabled` flag is reset when `accessibilityState.disabled` goes `true → undefined` (vs `true → false`). New Arch has been the only path since 0.82, so this is a refinement inside Fabric / the Android view manager rather than an arch switch. That also explains why this repo (pinned to 0.83.6) never caught it in CI. The fix is version-agnostic since it makes the transition an explicit `true ↔ false`.

PR #497, opened 2026-06-05
Implement support for CSS light-dark() in css.defineVars

efoken:
As suggested in the StyleX docs, modern browsers should use the `light-dark()` CSS function instead of theme-based overrides. This simplifies the code and removes the need to explicitly define dark mode themes.

This PR adds support for `light-dark()` in `css.defineVars`:

```ts
const tokens = css.defineVars({
  lightDarkColor: 'light-dark(blue, green)'
});
```

For React Native, this is functionally equivalent to defining separate values and resolving them via the `@media (prefers-color-scheme: dark)` Media Query.

In modern browsers, `light-dark()` also allows the color scheme to be controlled via the `color-scheme` CSS property, instead of relying on custom theme logic or duplicated Media Query-based definitions.

This aligns the implementation with the current StyleX recommendation and modern CSS color-scheme handling.

### DIFF
--- a/packages/react-strict-dom/src/native/css/customProperties.js
+++ b/packages/react-strict-dom/src/native/css/customProperties.js
@@ -40,6 +40,28 @@ export function stringContainsVariables(input: string): boolean {
   return input.includes('var(');
 }
 
+function stringContainsLightDark(input: string): boolean {
+  return input.includes('light-dark(');
+}
+
+const RE_LIGHT_DARK = /^light-dark\(\s*(.+?)\s*,\s*(.+?)\s*\)$/i;
+
+function resolveLightDarkValue(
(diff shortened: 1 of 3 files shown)


cmrn:
Thanks for working on this - I'm keen to see `light-dark()` support land in RSD.

One issue with the current approach: on web, `light-dark()` resolves against the `color-scheme` set at that point in the tree, not the media query. Resolving from `prefers-color-scheme` alone won't match web behaviour - e.g. a subtree with `color-scheme: dark` inside a light page resolves to the dark value on web, but this implementation would resolve it to light.

To get parity we'd need to track the `color-scheme` setting within the tree, similar to how CSS custom properties or inherited styles are handled today.

PR #499, opened 2026-06-08
[react-strict-animated] Fix incorrect error message for invalid mass in SpringAnimation

adityasingh2400:
## Summary

The `mass` validation in the `SpringAnimation` constructor throws the wrong error message. It reports a damping problem when an invalid mass is supplied, because the message was copy-pasted from the preceding damping check.

```js
if (this.#damping <= 0) {
  throw new Error('Damping value must be greater than 0');
}
if (this.#mass <= 0) {
  throw new Error('Damping value must be greater than 0'); // should be Mass
}
```

A user passing `mass: 0` (or a negative mass) gets told their damping value is wrong, which points them at the wrong config field. React Native's equivalent `SpringAnimation` throws `'Mass value must be greater than 0'` here.

## Fix

Use the correct message for the mass check.

## Test

Added `SpringAnimation-test.js` covering all three constructor validations (stiffness, damping, mass). The mass case fails on `main` (asserts `'Mass value must be greater than 0'`) and passes with the fix.

`flow check` and `eslint` are clean.

---

Reopening this. I closed it by accident during a bulk cleanup of my stale forks and cannot reopen the original because the fork was deleted, so this is the same branch restored. Sorry for the noise @MoOx.

### DIFF
--- a/packages/react-strict-animated/src/web/__tests__/SpringAnimation-test.js
+++ b/packages/react-strict-animated/src/web/__tests__/SpringAnimation-test.js
@@ -1,0 +1,30 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow strict-local
+ */
+
+import SpringAnimation from '../animations/SpringAnimation';
+
+describe('SpringAnimation (web)', () => {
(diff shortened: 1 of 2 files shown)


martinbooth:
If you could run `yarn format` or `npm run format` then I can merge this. Don't worry about the other failed actions, they are issue on main and I've (hopefully) fixed them now (so rebasing would be helpful too)

Thanks!

PR #501, opened 2026-06-12
RFC: Improvements to compat.native to support resolving vars correctly and access to the inherited text style

martinbooth:
Add a second argument to the react.compat callback which exposes StrictReactNativeMetaProps

StrictReactNativeMetaProps has:

# resolveStyleValue

A function that will resolve variables to their concrete values

```
resolveStyleValue('red')            // → 'red'        (no var → returned unchanged)
resolveStyleValue('10px')           // → '10px'       (no var → unchanged)
resolveStyleValue(vars.brand)       // → 'magenta'    (token resolved through the theme cascade)
resolveStyleValue('var(--brand)')   // → 'magenta'    (explicit var() resolved)
resolveStyleValue('var(--missing)') // → null         (unresolvable reference)
```

- Ancestor/ingerited theme tokens correctly resolve using this function
- Light/dark tokens resolve correctly
- Correctly resolves fallbacks, nested vars, non color vars, etc
- Avoid exposing any private internals

# inheritedTextStyle

Provides access to the text style that would be inherrited by child component

Importantly, it provide access to the color which is used on things like `svg` elements

# A more real world example:

```
import { compat } from 'react-strict-dom';
import { Path as RNPath } from 'react-native-svg';

function Path({ style, fill, stroke, ...rest }) {
  return (
    <compat.native style={style}>
      {(nativeProps, { inheritedTextStyle, resolveStyleValue }) => (
        <RNPath
          {...rest}
          // RSD doesn't know about SVG paint props, so resolve var()/tokens yourself:
          fill={fill != null ? resolveStyleValue(fill) : undefined}
          stroke={stroke != null ? resolveStyleValue(stroke) : undefined}
          // wire CSS `color` inheritance so `currentColor` works:
          color={inheritedTextStyle.color}
        />
      )}
    </compat.native>
  );
}
```

### DIFF
--- a/packages/react-strict-dom/src/native/compat.js
+++ b/packages/react-strict-dom/src/native/compat.js
@@ -8,6 +8,7 @@
  */
 
 import type { StrictProps } from '../types/StrictProps';
+import type { StrictReactNativeMetaProps } from '../types/renderer.native';
 
 import * as React from 'react';
 
@@ -31,7 +32,9 @@ import { createStrictDOMTextInputComponent as createStrictTextInput } from './mo
  *   aria-label="label"
  *   as="text"
  * >
- *   {(nativeProps: React.PropsOf<Text>)) => (
+ *   {(nativeProps: React.PropsOf<Text>, meta) => (
+ *     // `meta.inheritedTextStyle` and `meta.resolveStyleValue` allow resolving
+ *     // CSS inheritance and `var(...)` for non-RSD host components.
  *     <Text {...nativeProps} />
  *   )}
  * </compat.native>
@@ -42,7 +45,7 @@ const defaultProps = {};
 type StrictPropsOnlyCompat<T> = {
   ...StrictProps,
   as?: 'div' | 'img' | 'input' | 'span' | 'textarea',
-  children: (nativeProps: T) => React.Node
+  children: (nativeProps: T, meta: StrictReactNativeMetaProps) => React.Node
 };
 
 const StrictText = createStrictText('span', defaultProps) as $FlowFixMe;
--- a/packages/react-strict-dom/src/native/modules/createStrictDOMComponent.js
+++ b/packages/react-strict-dom/src/native/modules/createStrictDOMComponent.js
@@ -8,6 +8,7 @@
  */
 
 import type { ReactNativeProps } from '../../types/renderer.native';
+import type { StrictReactNativeMetaProps } from '../../types/renderer.native';
 import type { StrictProps as StrictPropsOriginal } from '../../types/StrictProps';
 
 import * as React from 'react';
@@ -15,7 +16,11 @@ import * as ReactNative from '../react-native';
 
 import { ProvideCustomProperties } from './ContextCustomProperties';
 import { ProvideDisplayInside, useDisplayInside } from './ContextDisplayInside';
-import { ProvideInheritedStyles } from './ContextInheritedStyles';
+import {
+  ProvideInheritedStyles,
+  useInheritedStyles
+} from './ContextInheritedStyles';
+import { flattenStyle } from './flattenStyle';
 import { TextString } from './TextString';
 import { errorMsg } from '../../shared/logUtils';
 import { useNativeProps } from './useNativeProps';
@@ -23,7 +28,9 @@ import { useStrictDOMElement } from './useStrictDOMElement';
 
 type StrictProps = Readonly<{
   ...StrictPropsOriginal,
-  children?: React.Node | ((ReactNativeProps) => React.Node)
+  children?:
+    | React.Node
+    | ((ReactNativeProps, StrictReactNativeMetaProps) => React.Node)
 }>;
 
 const AnimatedPressable = ReactNative.Animated.createAnimatedComponent(
@@ -54,15 +61,22 @@ export function createStrictDOMComponent<T, P extends StrictProps>(
      * Resolve global HTML and style props
      */
 
-    const { customProperties, nativeProps, inheritableStyle } = useNativeProps(
-      defaultProps,
-      props,
-      {
-        provideInheritableStyle,
-        withInheritedStyle: false,
-        withTextStyle: false
-      }
-    );
+    const {
+      customProperties,
+      resolveStyleValue,
+      nativeProps,
+      inheritableStyle
+    } = useNativeProps(defaultProps, props, {
+      provideInheritableStyle,
+      withInheritedStyle: false,
+      withTextStyle: false
+    });
+
+    const ancestorInheritedStyle = useInheritedStyles();
+    const inheritedTextStyle = flattenStyle([
+      ancestorInheritedStyle,
+      inheritableStyle
+    ]);
 
     if (
       nativeProps.onPress != null &&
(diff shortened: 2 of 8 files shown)


yaminyassin:
This direction looks right. I got this exact gap wrapping `react-native-svg` in my design system: 

```typescript
export const Icon = ({ children, size, color, viewBox = "0 0 24 24", fill = "none", ...props }) => (
  <compat.native {...props} as="span">
    {(nativeProps: SvgProps) => (
      <Svg {...nativeProps} width={size} height={size} viewBox={viewBox} fill={fill} color={color}>
        {children}
      </Svg>
    )}
  </compat.native>
);
```
We can't pass our css tokens straight to the svg, so we keep a separate map of resolved token values alongside the styles just for these cases. A `resolveStyleValue` callback would help here.

One thing that annoys me is having to break from the style creation for these custom values. Pulling values out and re-applying CSS is the part RSD usually handles. Normally we hand it a style and spread the result. 

What if we went on a different direction instead? we could have one field that takes the whole style and gives back resolved values to spread. I'm imagining something like this:

```typescript
function Path({ style, ...rest }) {
  return (
    <compat.native style={style}>
      {(nativeProps, { resolveHostStyle }) => (
        <RNPath {...rest} {...resolveHostStyle(style)} />
      )}
    </compat.native>
  );
}

// caller authors the paint in the style; RSD resolves it
const styles = css.create({
  mark: { color: "currentColor", "--fill": vars.brand, "--stroke": "currentColor" }
});

// or, with a passthrough:
const styles = css.create(
  { mark: { color: "currentColor", fill: vars.brand, stroke: "currentColor" } },
  { passthrough: ["fill", "stroke"] }
);
```
Having a prefix or a passthrough that allows unsupported styles to be passed and be resolved in `resolveHostStyle` function that resolves vars/tokens/dark-light and handles inheritedStyles without exposing this to the user. On the downside this would be forcing us to have weird rules around `css.create`.

what do you think?

martinbooth:
> This direction looks right. I got this exact gap wrapping `react-native-svg` in my design system:
> 
> ```ts
> export const Icon = ({ children, size, color, viewBox = "0 0 24 24", fill = "none", ...props }) => (
>   <compat.native {...props} as="span">
>     {(nativeProps: SvgProps) => (
>       <Svg {...nativeProps} width={size} height={size} viewBox={viewBox} fill={fill} color={color}>
>         {children}
>       </Svg>
>     )}
>   </compat.native>
> );
> ```
> 
> We can't pass our css tokens straight to the svg, so we keep a separate map of resolved token values alongside the styles just for these cases. A `resolveStyleValue` callback would help here.
> 
> One thing that annoys me is having to break from the style creation for these custom values. Pulling values out and re-applying CSS is the part RSD usually handles. Normally we hand it a style and spread the result.
> 
> What if we went on a different direction instead? we could have one field that takes the whole style and gives back resolved values to spread. I'm imagining something like this:
> 
> ```ts
> function Path({ style, ...rest }) {
>   return (
>     <compat.native style={style}>
>       {(nativeProps, { resolveHostStyle }) => (
>         <RNPath {...rest} {...resolveHostStyle(style)} />
>       )}
>     </compat.native>
>   );
> }
> 
> // caller authors the paint in the style; RSD resolves it
> const styles = css.create({
>   mark: { color: "currentColor", "--fill": vars.brand, "--stroke": "currentColor" }
> });
> 
> // or, with a passthrough:
> const styles = css.create(
>   { mark: { color: "currentColor", fill: vars.brand, stroke: "currentColor" } },
>   { passthrough: ["fill", "stroke"] }
> );
> ```
> 
> Having a prefix or a passthrough that allows unsupported styles to be passed and be resolved in `resolveHostStyle` function that resolves vars/tokens/dark-light and handles inheritedStyles without exposing this to the user. On the downside this would be forcing us to have weird rules around `css.create`.
> 
> what do you think?

I think there are two pieces of feedback here.. the ergonomics of `resolveStyleValue` and whether `css.create` can pass through unknown styles.

For `resolveStyleValue`. I'm trying to avoid making assumptions on how it might be used.. while spreading an object of styles might be one use case, I suspect forcing someone to construct an object with a single field as a style and then destructuring that field off again for the single style use-case is very unergonomic. It should be pretty straightforward to implement a`resolveHostStyle` helper which takes both a style and the `resolveStyleValue` function

For `css.create`. I think this is a separate discussion/RFC, but these proposals won't work. 

- `--fill: vars.brand` would actually already get passed through (-- is already not filtered), but this is a custom property definition (i.e. a css variable). On web, this would not get mapped to the fill css property as you probably want. While you could probably find a way of mapping it to the fill prop on RNPath today, that wouldn't produce xplat compatible code
- `{ passthrough: ["fill", "stroke"] }` would be a breaking change to stylex and also it wouldn't make sense for web

For now, I believe using svg props rather than trying to style svg with stylex is going to be the most practical short-term solution.

One final comment, using `as="span"` does have side effects (creating a text context, etc) so it's not ideal, but I understand it's currently the only way to expose the current text color.. something like this should provide a better way to get that color without the side effects

yaminyassin:
That distinction makes sense. I was mixing two things together. Agreed on `css.create` passthrough/custom props being a separate RFC. I was mostly trying to keep authors inside style creation.

The `resolveStyleValue` + `inheritedTextStyle` solves the current gap for me, especially removing the need for the `as="span"` workaround just to read inherited color. A small example in docs would be useful so people know this exists as well.

axyz:
great addition!

I started sometime ago looking into possible improvements to compat.native as well as part of https://github.com/react/react-strict-dom/discussions/479

There are 2 additional features that could be worth adding beside those in this PR:

- ability to have multiple style props (to avoid the nested compat.native anti-pattern
- ability to separate rsd "strict" props from the rest

some sample generated tests to illustrate:

```tsx
test('styleProps support', () => {
    const styles = css.create({
      container: {
        paddingInline: '2rem'
      },
      content: {
        fontSize: '1.5rem',
        color: 'blue'
      },
      columns: {
        gap: '1rem'
      }
    });

    let root;
    let capturedProps;
    act(() => {
      root = create(
        <compat.native
          style={styles.container}
          styleProps={{
            contentContainerStyle: styles.content,
            columnWrapperStyle: styles.columns
          }}
        >
          {(nativeProps) => {
            capturedProps = nativeProps;
            return <View {...nativeProps} />;
          }}
        </compat.native>
      );
    });

    expect(capturedProps).toBeDefined();
    expect(capturedProps.style).toEqual({
      paddingHorizontal: 32
    });
    expect(capturedProps.contentContainerStyle).toEqual({
      fontSize: 24,
      color: 'blue'
    });
    expect(capturedProps.columnWrapperStyle).toEqual({
      gap: 16
    });
    expect(root.toJSON()).toMatchSnapshot('styleProps');
  });

```

```tsx
test('partitions standard strict props and unknown passthrough props', () => {
    const props = {
      style: { color: 'blue' },
      onClick: () => {},
      'aria-label': 'My Button',
      'data-testid': 'test-btn',
      // Unknown/Native-only props
      numColumns: 3,
      onEndReached: () => {},
      customProp: 'value'
    };

    const { strictProps, passthroughProps } = splitStrictProps(props);

    expect(strictProps).toEqual({
      style: props.style,
      onClick: props.onClick,
      'aria-label': props['aria-label'],
      'data-testid': props['data-testid']
    });

    expect(passthroughProps).toEqual({
      numColumns: 3,
      onEndReached: props.onEndReached,
      customProp: 'value'
    });
  });
```

these additions should allow better interop with native components avoiding the nested compat.native and avoiding the error prone manual cherry-pick of props to pass to the native component vs the compat.native.

Let me know your thoughts and if you have different alternatives in mind. I also have an exploratory branch for a possible implementation to share in case.

PR #515, opened 2026-08-05
Call inherited host node methods on the host node in strict refs

### LINKED ISSUE #506: html.input ref .focus() does not work on native
## Describe the issue

Calling `.focus()` on a ref attached to `html.input` or `html.textarea` does nothing on native. The input is not focused and the keyboard does not appear. The workaround is to call `.getNativeRef().focus()` instead, which accesses the underlying React Native `TextInput` directly.

## Expected behavior

`ref.current.focus()` should focus the input and show the keyboard, the same way it works on web.

## Steps to reproduce

`react-strict-dom` version: `0.0.55`
Platform: iOS (React Native)

1. Create an `html.input` with a `ref`
2. Call `ref.current.focus()` programmatically (e.g. from a button press)
3. Observe: nothing happens, input is not focused

**Root cause:** PR #392 replaced explicit method proxying with `Object.create(node) + Object.defineProperties(strictRef, Object.getOwnPropertyDescriptors(node))`. The old code explicitly proxied `focus: proxy(node, 'focus')`. The new approach relies on `getOwnPropertyDescriptors(node)` to copy `focus` — but `focus` on a React Native `TextInput` ref is defined on the **prototype chain**, not as an own enumerable property, so it is not captured and silently disappears from `strictRef`.

**Workaround:**
```ts
// broken
ref.current?.focus();

// works
(ref.current as any)?.getNativeRef?.()?.focus?.();
```

## Test case

<details>
<summary>App.tsx — minimal reproduction</summary>

```tsx
import React, { useRef } from 'react';
import { css, html } from 'react-strict-dom';

export default function App() {
  const inputRef = useRef<HTMLInputElement>(null);

  return (
    <html.div data-layoutconformance="strict" style={styles.container}>
      <html.input ref={inputRef} placeholder="Target input" style={styles.input} />
      <html.button onClick={() => inputRef.current?.focus()} style={styles.button}>
        ref.current.focus() — broken
      </html.button>
      <html.button
        onClick={() => (inputRef.current as any)?.getNativeRef?.()?.focus?.()}
        style={styles.button}
      >
        getNativeRef().focus() — works
      </html.button>
      <html.input placeholder="Tap here to reset focus" style={styles.input} />
    </html.div>
  );
}

const styles = css.create({
  container: {
    paddingTop: 60,
    paddingLeft: 24,
    paddingRight: 24,
    gap: 16,
  },
  input: {
    borderWidth: 1,
    borderStyle: 'solid',
    borderColor: '#999',
    borderRadius: 4,
    padding: 8,
  },
  button: {
    padding: 12,
    borderRadius: 4,
    backgroundColor: '#e0e0e0',
    ':active': { backgroundColor: '#b0b0b0' },
  },
});
```

</details>

## Additional comments

**Suggested fix** — explicitly forward `focus` and `blur` for `input`/`textarea` strictRefs in [`useStrictDOMElement.js`](https://github.com/react/react-strict-dom/blob/main/packages/react-strict-dom/src/native/modules/useStrictDOMElement.js#L135), similar to how `setSelectionRange` is currently handled:

```js
const focus = node?.focus;
if (focus) {
  Object.defineProperty(strictRef, 'focus', {
    value: (...args) => focus.call(node, ...args),
    configurable: true,
    writable: true,
  });
}
```

adityasingh2400:
Fixes #506.

`getOrCreateStrictRef` builds the strict ref with `Object.create(node)`, so a method that lives on the host node prototype chain is still reachable from the wrapper but gets invoked with the wrapper as its receiver. React Native host node methods read internals off `this`, and the wrapper does not carry those internals, so `ref.current.focus()` and `ref.current.blur()` silently do nothing on native while `getNativeRef().focus()` still works. Reads keep working because own properties of the node resolve through the prototype chain, which is why only the imperative methods broke.

This walks the node's prototype chain once per node and re-exposes each inherited method as an own property that calls through to the node, so the receiver stays correct. Own methods of the node are left alone since React Native assigns those as closures over the node, and accessors are left on the prototype chain since they read plain instance fields the wrapper already inherits. The binding runs before the strict-dom overrides are defined so the viewport-scaled `getBoundingClientRect` still wins.

Added a regression test in `html-refs-test.native.js` that mounts an `html.input` over a host node mock whose `focus` and `blur` live on the prototype, then asserts the receiver is the host node. It fails on main and passes with this change. Ran the full package suite (19 suites, 897 tests), `flow check`, eslint, and prettier, all clean.

### DIFF
--- a/packages/react-strict-dom/CHANGELOG.md
+++ b/packages/react-strict-dom/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 * [Native] React Strict DOM and React Strict Animated now require `react-native >=0.82.0` for native builds that use RN DOM Node APIs.
 
+### Fixes
+
+* [Native] Call inherited host node methods, such as `focus` and `blur`, on the host node so imperative ref APIs work again.
+
 ### Internal
 
 * [Native] `useStrictDOMElement` now wraps the RN host node with `Object.create(node)` and defines only strict-dom overrides as own properties.
--- a/packages/react-strict-dom/src/native/modules/useStrictDOMElement.js
+++ b/packages/react-strict-dom/src/native/modules/useStrictDOMElement.js
@@ -49,6 +49,8 @@ type StrictRefTarget = {
 const memoizedStrictRefs: WeakMap<HostInstance, StrictRefTarget> =
   new WeakMap();
 
+const objectPrototype: interface {} | null = Object.getPrototypeOf({});
+
 const lengthPropertySet: ReadonlySet<string> = new Set([
   'clientHeight',
   'clientLeft',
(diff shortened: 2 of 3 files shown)


martinbooth:
Thank you for fixing this!

PR #518, opened 2026-09-04
Disable Android Image fadeDuration on native html.img

### LINKED ISSUE #461: Image fadeDuration property
react-native has an android only property `fadeDuration` that adds a fade-in effect to images while they load.

This is set by default to `300ms` leading to an unexpected inconsistency across platforms.

the issue was discussed here: https://github.com/facebook/react-native/issues/10194

where was decided to add a prop to control it, but not removing the default value to avoid breaking changes.

---

I'm wondering what is the position on the topic from the react-strict-dom POV. Should the `createStrictImage` force it to `0` and let the optional animation be handled by css transition polyfills?

If not, how can the behavior be overridden? In react-native the fadeDuration property can be used to reset to 0 when needed, but not being a valid web API is not allowed in `html.img` (as well as possibly creating conflicts with transitions and/or animations)

bencripps:
## Summary
- Fixes #461 by setting RN `Image` `fadeDuration` to `0` on native `html.img`.
- Android otherwise fades images in over 300ms by default, which is not a web behavior and cannot be disabled through a web-valid prop.
- CSS transitions can still animate opacity. `compat.native` callers can override `fadeDuration` after spreading `nativeProps`.

## Test plan
- [x] Added a native unit test asserting `html.img` passes `fadeDuration: 0`
- [x] Updated native image snapshots
- [ ] Maintainer: confirm Android `html.img` no longer fades in on load

### DIFF
--- a/packages/react-strict-dom/src/native/modules/createStrictDOMImageComponent.js
+++ b/packages/react-strict-dom/src/native/modules/createStrictDOMImageComponent.js
@@ -85,6 +85,10 @@ function applyImageProps(
 
   // Component-specific props
 
+  // RN Image fades in over 300ms on Android by default. That is not a web
+  // behavior, so disable it. CSS transitions can still animate opacity.
+  nativeProps.fadeDuration = 0;
+
   nativeProps.ref = elementRef;
 }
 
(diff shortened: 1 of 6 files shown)


axyz (approved):

bencripps:
any chance we could get this merged? thanks!

